PR #26, opened 2022-02-03
The `/api/source_code`-endpoint now references a specific commit instead of the base

### LINKED ISSUE #25: The `/api/source_code`-endpoint should reference a specific commit
The Commit data can be supplied via an environment variable given in the docker build.
If no variable is given the plain reposittory should be shown

CommanderStorm:
The Commit data can be supplied via an environment variable given in the docker build.
If no variable is given the plain reposittory should be shown

resolves #25

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -5,7 +5,6 @@ use std::fs;
 
 use actix_cors::Cors;
 use actix_web::{get, http, middleware, web, App, HttpRequest, HttpResponse, HttpServer, Result};
-use serde_json::to_string;
 use structopt::StructOpt;
 
 mod feedback;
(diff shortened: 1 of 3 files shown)


server/src/main.rs
```diff
@@ -1,9 +1,11 @@
 #[macro_use]
 extern crate lazy_static;
+
 use std::fs;
 
 use actix_cors::Cors;
 use actix_web::{get, http, middleware, web, App, HttpRequest, HttpResponse, HttpServer, Result};
+use serde_json::to_string;
```
octycs, line 8: This appears to not to be used and triggers an `unused import` warning

octycs (commented):
I get one `unused import` warning, but apart from that I think everything is fine

server/src/main.rs
CommanderStorm, line 8:
```suggestion
```

PR #29, opened 2022-02-04
Minor restyling of the search results

CommanderStorm:
pushed the runtime and feedback up next to what we searched for (like…google)

Before: 
Screenshot_20220204_220127

After:
Screenshot_20220204_220040

### DIFF
--- a/webclient/src/views/search/view-search.js
+++ b/webclient/src/views/search/view-search.js
@@ -39,11 +39,7 @@ navigatum.registerView('search', {
         loadSearchData: function(query, data) {
             this.search_data = data;
             this.query = query;
-            var search= document.getElementById("search")
-            if (search.value.length === 0) {
-                // /search/:querry is called from an internal url and thus the search url is not set
-                search.value = query;
-            }
+            navigatum.app.search.query = query;
             navigatum.setTitle('${{ _.view_search.search_for }}$ "' + query + '"');
             
             // Currently borrowing this functionality from autocomplete.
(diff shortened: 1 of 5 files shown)


CommanderStorm:
What is your oppinion on leaving the seach result in the seach bar?
I noticed that you made this configurable, but then decided for the current option

octycs:
Probably makes sense there. I configured to remove the query when you select an element from the autocompletion, because I consider a search "finished" then.

webclient/src/views/search/view-search.js
```diff
@@ -39,7 +39,11 @@ navigatum.registerView('search', {
         loadSearchData: function(query, data) {
             this.search_data = data;
             this.query = query;
-            
+            var search= document.getElementById("search")
+            if (search.value.length === 0) {
+                // /search/:querry is called from an internal url and thus the search url is not set
+                search.value = query;
+            }
```
CommanderStorm, line 46: What do you think about this? 
is this nessesary?
Through the bug in the ngnix config I am unable to check, if this could work

CommanderStorm:
updated UI-sample
image

webclient/src/views/search/view-search.js
octycs: It seems not. When I use hash based navigation, vue-router shows a 404 for `search/` without a query.
Instead of using the `document.getElementById` approach you can however set the query using vue.js directly. Changing the element works as well, but this would be more consistent with the rest of the app.

```suggestion
            navigatum.app.search.query = query;
```

CommanderStorm: Thanks :)

PR #31, opened 2022-02-04
Support for custom maps + StudiTUM Garching maps floor 0-3

octycs:
This PR adds the possibility to add custom maps that appear in the Roomfinder map lists and includes some maps for the StudiTUM Garching along with coordinates.

ToDos:
- [x] Unify and document data format changes (especially ids are now mixed between str and int)


CommanderStorm:
what is the status of this PR? 
I dont see where the id is a str and int..

octycs:
This PR introduces string map ids (e.g. `"5532_3"`), which have so far been ints without any meaning (taken from the original RoomFinder). I tend to prefer strings here, because buildings ids etc are also strings, so we can turn them into a common format (e.g. `"<entry id>_<floor>"`).

What is still to do is converting the RoomFinder map ids into that format. I think I'll do a simple solution here for now, so that e.g. `87` will just be converted into a string `"rf87"`.

The second thing to do is applying the switch to webp. It's not significant in absolute terms (50kB PNG -> 25kB lossless WebP, interestingly lossy WebP gives me 28kB), but still an optimization.

I'll do that today or in the next days.

octycs:
This should be ready now

data/processors/maps.py
```diff
@@ -284,9 +302,51 @@ def build_roomfinder_maps(data):
...
+                **base_data
+            }
+
+    return maps_out
+        
+    
+    
+    
+    
+
```
CommanderStorm, line 352:
```suggestion

```

data/processors/maps.py
```diff
@@ -284,9 +302,51 @@ def build_roomfinder_maps(data):
...
+def _load_custom_maps():
+    """ Load the custom maps like Roomfinder maps """
+    with open("sources/45_custom-maps.yaml") as f:
+        custom_maps = yaml.safe_load(f.read())
+        
+    # Convert into the format used by maps_roomfinder.json:
+    maps_out = {}
+    for map_group in custom_maps:
+        base_data = {
+            "source": map_group["props"].get("source", "NavigaTUM"),
```
CommanderStorm, line 324: Should we credit `NavigaTUM-Contributors` instead?

data/processors/maps.py
```diff
@@ -284,9 +302,51 @@ def build_roomfinder_maps(data):
...
+            "latlonbox": {
+                "north":    map_group["props"]["north"],
+                "east":     map_group["props"]["east"],
+                "west":     map_group["props"]["west"],
+                "south":    map_group["props"]["south"],
+                "rotation": map_group["props"]["rotation"],
+            }
+        }
+        for sub_map in map_group["maps"]:
+            img = Image.open("sources/img/maps/roomfinder/webp/" + sub_map["file"])
```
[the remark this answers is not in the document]
octycs: Yes, I'm currently working on overlay maps for the interactive map

data/processors/maps.py
```diff
@@ -225,21 +241,23 @@ def assign_roomfinder_maps(data):
...
+    # Read the Roomfinder and custom maps
     with open("external/maps_roomfinder.json") as f:
         maps_list = json.load(f)
+    custom_maps = _load_custom_maps()
     
-    # For each map, we calculate the boundaries in UTM, because 
+    # For each map, we calculate the boundaries in UTM beforehand
     maps = {}
-    for m in maps_list:
+    for m in maps_list + list(custom_maps.values()):
```
CommanderStorm, line 251: `_load_custom_maps()` should return the format we need up here. 
Thus I think we should move this conversion into the more proper method.

data/processors/maps.py
```diff
@@ -128,6 +130,9 @@ def assign_roomfinder_maps(data):
     # Read the Roomfinder maps
     with open("external/maps_roomfinder.json") as f:
         maps_list = json.load(f)
+        
+    # There are also Roomfinder-like custom maps, that we assign here
+    custom_maps = _load_custom_maps()
```
CommanderStorm, line 135: I dont love, how this is 51 lines away from its first usage.
This is however not a thing I'd expect you to fix, since this method would be refactored for this to be fixed. (that would have to be another PR)

CommanderStorm (changes requested):
I dont like, how the folder-layout (e.g. maps shouldn't be in different folders) is, but that would be work for another PR.

- [ ] please add the custom maps-image folder to the dockerfile. Otherwise those files wont be availibe in prod


I left some inline-comments

data/processors/maps.py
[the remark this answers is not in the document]
CommanderStorm: Fancy.

data/processors/maps.py
CommanderStorm:
```suggestion
            "source": map_group["props"].get("source", "NavigaTUM-Contributors"),
```

data/processors/maps.py
octycs: The whole functions are quite large actually, which is certainly not ideal. The reason I put this there is becaue I wanted to group map loading code.

octycs:
I understand that issue with different directories, which makes it more tedious, but I'd prefer to keep external and internal data separate in some way to make proper attribution.

data/processors/maps.py
CommanderStorm: I know, that is something I would have done too, but we should keep this in mind, before adding to this function in the future

CommanderStorm:
I think, that internal and external are already mixed.
We do include images from external sources in the sources folder...

CommanderStorm:
Changing the folder-structure is a different PR however (and probably requires your maps overlay feature, since I dont want to cause you extra work..)

octycs:
> I dont like, how the folder-layout (e.g. maps shouldn't be in different folders) is, but that would be work for another PR.
> 
>     * [ ]  please add the custom maps-image folder to the dockerfile. Otherwise those files wont be availibe in prod
> 
> 
> I left some inline-comments

This should actually be already correct (tested in a local build). I just noticed that this is another reason for that folder structure with the `webp` subdir, because the `COPY sources/img/. output/` will already copy them to `output/maps/roomfinder/webp`.

CommanderStorm (approved):

PR #32, opened 2022-02-06
Added a custom maps style

### LINKED ISSUE #8: Alternative maps-style
We should consider including a different maps design:
instead of the normal OSM-view, we could include a custom style.

A website doing this is the [rallye-tool](https://rallye.mpi.fs.tum.de/ratings/map/):
image

The code for this is availible at https://github.com/FSTUM/rallyetool-v2/blob/main/templates/registration/map.html

CommanderStorm:
- [x] integrated our alternative map style
- [x] changed to a custom (only for this project) style.json
- [x] integrated into the build system

# Discussion
The current implementation has slightly worse performance, due to using https://github.com/mapbox/mapbox-gl-leaflet under the hood.
This is because:

> Slower performances: When using mapbox-gl-leaflet, mapbox-gl-js is set as not interactive. Leaflet receives the touch/mouse events and updates the mapbox-gl-js map behind the scenes. Because mapbox-gl-js doesn't redraw as fast as Leaflet, the map can seem slower.

Alternatives I have considered:
- using a custom rasterisation service. This is costwise not an option for us, as I dont want to pay as much as for example MapTiler charges for serverside rasterisation
- using a tilejson compatible api. Not an option, because the one I found only has guides for vue-3 (we currently use vue-2) and the libarys I found which support vue-2 have other drawbacks. There may be one which I have not found, but currently this is not the case

resolves #8 

# Screenshots
before:
image

after:
image


octycs:
I like that these vector tiles are a lot sharper on continuous zooming and allow to increase the zoom level (which might be interesting if we overlay indoor maps some time in the future). From some tests I did, the rendering performance seems sufficient, only initial data loading appear slower (but this might have different reasons).

I could only do a little research, but I'd appreciate if we could host the tiles ourselves, which should be possible because we only need to cover a small area for the campus locations (with maybe a hybrid approach to load OSM tiles when zooming out far). Some interesting tools I found are:
- [tilemaker](https://tilemaker.org/) for .mbtiles generation (only needed once in a while)
- [mbtileserver](https://github.com/consbio/mbtileserver) for serving

There are a lot more similar projects (just to dump them here: [GeoServer](http://geoserver.org/), [baremaps](https://github.com/baremaps/baremaps), [tegola](https://github.com/go-spatial/tegola)). Unfortunately I don't have the time within the next month to look into this, so this is rather out of scope for this issue.

However one thing, is it possible to enable more labels (especially street names and subway/bus stations) on lower zoom levels?

CommanderStorm:
Yes, that is very much possible.
They are currently disabled, because I did not create a custom look for this project and just reused a look I created in the past. ;)

About creating custom tilesets I would like to hold off on, as this seems to require a lot of storage and my server upgrade plan has me on 40GB until the summer (after that I will upgrade and this should not be an issue anymore) 
Thanks for sharing your search results. I must have had some search tearms guiding me in the B2B-realm, which is why I thought of this as infeasable.

CommanderStorm:
> However one thing, is it possible to enable more labels (especially street names and subway/bus stations) on lower zoom levels?

What do you think about this current design:
https://api.mapbox.com/styles/v1/commanderstorm/ckzdc14en003m14l9l8iqwotq.html?title=copy&access_token=pk.eyJ1IjoiY29tbWFuZGVyc3Rvcm0iLCJhIjoiY2t6ZGJyNDBoMDU2ZzJvcGN2eTg2cWtxaSJ9.PY6Drc3tYHGqSy0UVmVnCg&zoomwheel=true&fresh=true#16.78/48.264624/11.670726

CommanderStorm:
> I like that these vector tiles are a lot sharper on continuous zooming and allow to increase the zoom level (which might be interesting if we overlay indoor maps some time in the future). From some tests I did, the rendering performance seems sufficient, only initial data loading appear slower (but this might have different reasons).

Initial loading performance seems fine (after not making three different requests).
The performance-issue can be provoked by circeling arround the screen with your cursor while dragging the map, but I suspect not a lot of people will notice this.

octycs:
I just tried it out locally and I'm getting a weird bug:
220208_bug
(Ignore the low GIF quality)

Have you encountered this already (Firefox 96.0.3)?

The console shows a lot of warnings and one error `The operation is insecure.` pointing to a `texImage2D` call.
image

There appears to be a relation to CORS, but all mapbox requests allow any origin, and I'm not sure whether this is the cause of the error. However I don't get it in the preview link you sent above.

octycs:
Going back to mapbox-gl to 2.4.1 solves this issue. This version doesn't have these nice zoom-fading effects, but for now that's probably the only option unless we patch mapbox-gl

CommanderStorm:
Honestly, I have had a really really hard time finding a test-case in which I can reporoduce this issue.  
The testcase is: use firefox, instead of chrome. 
Every map is imo unusable.

Solutions:
- a proper map tiling service (as mentioned above, coming in the summer when my current vps contract runs out)
- or matbox without improper hacks like the one `mapbox-gl-leaflet` really is..

Note that 
> Going back to mapbox-gl to 2.4.1 solves this issue
does not resolve this testcase.
If it were only chrome (where I was really really frustrated, that I was unable to  reproduce this bug)

CommanderStorm:
what is your environment?
I use 
```
Version 97.0b9 (64-Bit)
Mozilla Firefox Developer Edition for Manjaro Linux
Manjaro - 1.0
```

octycs:
I'm just one version behind on the latest stable:
```
96.0.3 (64-bit)
Mozilla Firefox for Manjaro Linux
Manjaro - 1.0
```

CommanderStorm:
That seems to not be the factor in wich our configuration differs.
I still am able to reproduce this issue on your version

CommanderStorm:
todos for tomorrow:
- make markerIcon use the right asset (not the placeholder)
- make gulp minify to .min files

CommanderStorm:
Things for other PRs are:
- figure out map-localisation
- figgure out how to restyle `GeolocateControl` and possibly add a sentence in the privacy statement, that `GeolocateControl` does not transmit location to the server

webclient/config-local.js
```diff
@@ -6,8 +6,8 @@ var config = {
     // hash-based navigation in development builds.
     app_prefix: "",
     // Prefix for 'cdn' content, e.g. images.
-    cdn_prefix: "/cdn/",
+    cdn_prefix: "https://roomapi.tum.sexy/cdn/",
     // Prefix for API requests
-    api_prefix: "http://localhost:8080/api/",
+    api_prefix: "https://roomapi.tum.sexy/api/",
```
octycs, line 11: It's probably a matter of preference, whether the local build should use the local server by default or the public one. I'd prefer using the local one by default, but I'm also okay changing this. However in this case we need to update the Readme then

webclient/src/modules/interactive-map.js
```diff
@@ -1,51 +1,67 @@
...
+            const map= new mapboxgl.Map({
+                container: container_id,
+
+                // create the gl context with MSAA antialiasing, so custom layers are antialiasing.
+                // slower, but prettier and therefore worth it for our use case
+                antialias: true,
+
+                //preview of the following style is available at
+                // https://api.mapbox.com/styles/v1/commanderstorm/ckzdc14en003m14l9l8iqwotq.html?title=copy&access_token=pk.eyJ1IjoiY29tbWFuZGVyc3Rvcm0iLCJhIjoiY2t6ZGJyNDBoMDU2ZzJvcGN2eTg2cWtxaSJ9.PY6Drc3tYHGqSy0UVmVnCg&zoomwheel=true&fresh=true#16.78/48.264624/11.670726
+                style: 'mapbox://styles/commanderstorm/ckzdc14en003m14l9l8iqwotq',
```
octycs, line 52:
```suggestion
                style: 'mapbox://styles/commanderstorm/ckzdc14en003m14l9l8iqwotq?optimize=true',
                
                center: [11.5748, 48.1400],  // Approx Munich
                zoom: 11,  // Zoomed out so that the whole city is visible
```

This sets the default view to approx Munich (zoomed out), so that one still gets an idea where that location is relative to/within Munich, but it doesn't have this long world zoom-in which requests a lot of data not really relevant.

Also, there appears to be an option to remove unused features and improve perfomance (https://docs.mapbox.com/help/troubleshooting/mapbox-gl-js-performance/#remove-unused-features)

webclient/src/views/view/view-view.js
```diff
@@ -168,32 +168,33 @@ navigatum.registerView('view', {
...
-                    });
-
-                    L.marker([coords.lat, coords.lon], {icon: c.icon}).addTo(map)
+                    marker = c.initMarker();
+                    _this.map.interactive.marker = marker;
+                    const coords = _this.view_data.coords;
+                    marker.setLngLat([coords.lon, coords.lat]).addTo(map);
+                    // Use 16 as default zoom for now, TODO: Compute
+                    if (from_map === "interactive"){
+                        map.flyTo({center: [coords.lon, coords.lat], zoom: 16, duration: 5000});
```
octycs, line 192:
```suggestion
                        map.flyTo({
                            center: [coords.lon, coords.lat],
                            zoom: 16,
                            speed: 1,
                            maxDuration: 2000
                        });
```

To be honest I think we need to reduce the fly time. It looks nice on fast devices, but introduces quite some lag on slower mobile devices.

octycs (commented):
The Mapbox Map certainly feels more interactive than the current one :+1: 

It doesn't load as quickly as Leaflet with static image maps for me (which is nearly instant), but I think this is sufficient for now. What we might want to look into some time is using WebWorkers for (pre)caching: https://docs.mapbox.com/mapbox-gl-js/api/properties/#prewarm

Also, we could try some time using the Static Images API (It might be possible to use https://github.com/jperelli/osm-static-maps for a self-hosted version): https://docs.mapbox.com/help/tutorials/improve-perceived-performance-with-static/

webclient/config-local.js
CommanderStorm: Oops.. that file should not have been commited

CommanderStorm:
@octycs Can this be merged/ do you have further comments?

octycs:
Thank you for creating the issue. I think this is ready to be merged.

PR #36, opened 2022-02-07
Added site and campus tags to two things in the areatree

CommanderStorm:
This is going though an PR, because I don't know if I have to modify other things

### DIFF
--- a/data/sources/00_areatree
+++ b/data/sources/00_areatree
@@ -457,7 +457,7 @@
       5142:Parkhaus P1:
       -5143:Versorgungsstation:
       5160:UCN-Testanlage (Container):
-    52:FRM II Forschungsreaktor Neutronenquelle und Heinz Maier-Leibnitz Zentrum (MLZ):frm-2[site]
+    52:FRM II Forschungsreaktor Neutronenquelle und Heinz Maier-Leibnitz Zentrum (MLZ):frm-2[area]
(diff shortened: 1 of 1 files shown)


data/sources/00_areatree
```diff
@@ -457,7 +457,7 @@
       5142:Parkhaus P1:
       -5143:Versorgungsstation:
       5160:UCN-Testanlage (Container):
-    52:FRM II Forschungsreaktor Neutronenquelle und Heinz Maier-Leibnitz Zentrum (MLZ):frm-2
+    52:FRM II Forschungsreaktor Neutronenquelle und Heinz Maier-Leibnitz Zentrum (MLZ):frm-2[site]
```
octycs, line 460:
```suggestion
    52:FRM II Forschungsreaktor Neutronenquelle und Heinz Maier-Leibnitz Zentrum (MLZ):frm-2[area]
```

octycs (approved):
The difference between "campus", "site" and "area" is not really well defined to be honest ("site" approx. means "Standort", a location smaller than a campus, and "area"s are so far parts of a campus).

That's why I would change it for the FRM II to an "area" of the Garching Campus, but Taufkirchen/Ottobrunn is definitely a campus, not an area (the default value) :+1:

CommanderStorm:
😅 I thought area is only for outdoor-areas (it is only used for an outdoor area before)

PR #38, opened 2022-02-08
moved all static vendors into npm for better dependency management

CommanderStorm:
This enables dependabot to pick up new releases and propose PRs for dependency updates.


octycs (approved):
Seems to work and makes sense :+1:

PR #43, opened 2022-02-18
SEO-Optimization

### LINKED ISSUE #13: SEO-Optimization
We should have a look at common SEO techniques (https://moz.com/learn/seo/on-site-seo).
- meta 
   - [x] descriptions
   - social tags (only static tags, see #45)
     - [x] Facebook-opengraph 
     - [x] twitter
   - Schema.org data (for more shemas see #46 )
     - [x] https://developers.google.com/search/docs/advanced/structured-data/breadcrumb#rdfa
- deployment
  - [x] look info serving images in modern formats like webp, JPEG 2000, ...  
        see https://web.dev/uses-webp-images/?utm_source=lighthouse&utm_medium=devtools
  - [x] look at the caching policy for images
- accesibility
  - [x] better image alt tags
  - [x] make shure buttons have acessile names  
         see https://web.dev/button-name/?utm_source=lighthouse&utm_medium=devtools
  - [x] have apropriatly sized touch targets on mobile:
    > Interactive elements like buttons and links should be large enough (48x48px), and have enough space around them, to be easy enough to tap without overlapping onto other elements see https://web.dev/tap-targets/?utm_source=lighthouse&utm_medium=devtools
- ~Image natural dimensions should be proportional to the display size and the pixel ratio to maximize image clarity see https://web.dev/serve-responsive-images/?utm_source=lighthouse&utm_medium=devtools some of our images have a lower resolution~

CommanderStorm:
This PR tries to improve our SE-Rating, the convergence rate and other things that might boost the projects popularity
resolves #13


CommanderStorm:
caveats:
- view building pages layout has slightly restyled
- the footer has been restyled
- I have added image-caching. The current expires value is set to 1 day.
  - Reasonging: images dont change so often and this may improve our search rating
  - We should look into https://www.npmjs.com/package/gulp-cache-bust or others to evaluate whether they are would be good aditions to the webserver (with this we could increase the cached resources and the cache durations)

webclient/src/index.html
```diff
@@ -1,10 +1,27 @@
...
+    <meta name="twitter:card" content="summary_large_image">
+    <meta property="og:type" content="website"/>
+    <meta property="og:title" content="NavigaTUM" />
+    <meta property="og:url" content="https://roomapi.tum.sexy">
+    <meta property="og:description" content="Navigate around TUM with excellence. This is a website to find rooms, buildings and other places at TUM easier than the current RoomFinders." />
+    <meta property="og:image" content="<!-- @echo app_prefix -->assets/logo-card.png" />
+    <meta property="og:image:width" content="1200" />
+    <meta property="og:image:height" content="630" />
+    <meta property="og:image:mime" content="image/png" />
+    <meta property="og:site_name" content="TUM.sexy" />
```
octycs, line 23:
```suggestion
    <meta property="og:site_name" content="NavigaTUM" />
```

I think this makes more sense on a social link card.

webclient/src/index.html
```diff
@@ -1,10 +1,27 @@
...
+<html lang="${{_lang_}}$" prefix="og: https://ogp.me/ns#">
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    
+
+    <meta name="description" content="Navigate around TUM with excellence. This is a website to find rooms, buildings and other places at TUM easier than the current RoomFinders.">
+    <meta name="author" content="@octycs and @CommanderStorm">
+    <meta name="copyright" content="GNU General Public License v3.0. Some images have other licences.">
```
octycs, line 10:
```suggestion
    <meta name="copyright" content="GNU General Public License v3.0. Images may be licensed differently.">
```

webclient/src/index.html
CommanderStorm: [ogp.me](https://ogp.me/) states:
> og:site_name - If your object is part of a larger web site, the name which should be displayed for the overall site. e.g., "IMDb".

I think, that TUM.sexy does fit this discription more, or do you think differently? I dont have a strong opinion on this

CommanderStorm:
> Thank you for your efforts, I'll look into it soon. Please be patient, I don't have a lot of time at the moment, so the suggested changes only concern the metadata at the moment.

That is to be assumed, when you open a PR directly in the exam-phase :)

> WebP definitely makes sense for all browsers, but I suspect we need to have a non-webp fallback, because Mac OS X only introduced it on Mac OS X 11 / Safari 14. I only have some statistics from a project a few months ago, but it appears there are still people using Mac OS X 10 and/or Safari 12 and 13. We could do it client-side by using `/webp/` or `/jpg/` etc., or maybe include a polyfill such as [webp-hero](https://github.com/chase-moskal/webp-hero) (I did only a quick research).

Fucking macOS… I could not find a statistic on this and assumed, that most people have upgraded, since it has been nearly half a year (it launched in November, could your numbers be biased?) since this launched.
I think, this can be safely ignored. At some point we have to draw a line and the population in the legacy macOS bucket seems to be not that large if I am honest.

> Also, I'm not sure, did you intend the styling changes? Honestly I'd prefer to reduce the font-size in the footer again.

Yes, the styling changes are intentional.
We can compromise and use a smaller font on larger devices.
The change was motivated, because Google told me, that the touch targets on mobile are too small…

CommanderStorm:
I had this polyfill tested by a friend with an antique macbook. It should work.

webclient/src/index.html
```diff
@@ -278,7 +318,7 @@
 
     <!-- Feedback modal -->
     <div class="modal" id="feedback-modal">
-      <a id="feedback-overlay" class="modal-overlay" aria-label="Close"></a>
+      <button id="feedback-overlay" class="modal-overlay" aria-label="Close"></button>
```
octycs, line 321:
```suggestion
      <div id="feedback-overlay" class="modal-overlay"></div>
```
This is actually not really a button, it's just the transparent background overlay that you can click on to make the dialog close. If it irritates Google I'd suggest using a `div` here, since a `button` breaks in Firefox for me.

Also, I'd remove the `aria-label` on this one then, because there is already a second (and real) close button below, and for screenreaders two buttons with the same text are probably not helpful.

webclient/src/views/view/view-view.scss
```diff
@@ -196,24 +194,55 @@
         }
     }
 
+    .menu {
+        padding: 0;
+        box-shadow: none;
+
+
```
octycs, line 201:
```suggestion

```

webclient/src/views/view/view-view.scss
```diff
@@ -196,24 +194,55 @@
...
+        .menu-item label,
+        .menu-item button {
+            cursor: pointer;
+            user-select: none;
+        }
+    }
+
     #rooms-overview {
+
+
```
octycs, line 218:
```suggestion
        #rooms-overview-select .menu-item {
            padding: 0;
            
            & .icon-arrow-right {
                margin-right: 4px;
            }
        }
```

This looks slightly better in my opinion. What do you think?

Before:
image

Now:
image

octycs (approved):
I did some changes to the layout again. Some just concern minor spacing correction, but I also changed the footer layout, so that it looks similar to the current on desktop, but uses the new layout on mobile.

If these changes are okay for you, I think this can be merged. It was honestly a bit hard to keep track of each individual changes in such a large PR.

CommanderStorm:
Yes, Looks good to me..
I should have broken this PR into smaller PRs honestly

PR #53, opened 2022-03-07
Made the webclient accept the Accept_language header

### LINKED ISSUE #47: [Feature] Set the default language using the `Accept-Language` header
Currently, the default language is German irrespective of the `Accept-Language` sent by the browser, because it is only determined by the `lang` cookie.

Configure Nginx to choose the language using the `Accept-Language` header if no cookie is set.

**Additional context**

CommanderStorm:
I am sorry this seems to be the "prettyest" way to get this done. There is a module to do this, but it essentially also does regexes (but a bit more efficiently since it can actually parse the content). The plugin however seems to be abandoned and installing it is not documented in a working way.

We might want to clean up the regex used for the`$ACCEPT_LANG` detection. Not from a performance but from a readability standpoint. This would make us less compliant to the standard, but this would imo still be acceptable.
resolves #47

### DIFF
--- a/webclient/nginx.conf
+++ b/webclient/nginx.conf
@@ -14,6 +14,40 @@ http {
   access_log  /var/log/nginx/access.log  main;
   sendfile        on;
   keepalive_timeout  65;
+
+  # theme detection
+  map $cookie_theme $THEME {
+    default "light";
+
+    "dark" "dark";
+    "light" "light";
+  }
+
+  # language detection. The accept_language header sucks.
+  # This map will only be used if $cookie_lang is unset
+  map $http_accept_language $ACCEPT_LANG {
+    default "de";
+
+    "~*.*de;.*en.*" "de";
+    "~*.*en;.*de.*" "en";
+    "~*.*de,.*en.*" "de";
+    "~*.*en,.*de.*" "en";
+    "~*.*de-.*en.*" "de";
+    "~*.*en-.*de.*" "en";
+    "~*.*de;.*" "de";
+    "~*.*en;.*" "en";
+    "~*.*de,.*" "de";
+    "~*.*en,.*" "en";
+    "~*.*de.*" "de";
+    "~*.*en.*" "en";
+  }
+
+  map $cookie_lang $LANG {
+    default "${ACCEPT_LANG}";
+    "en" "en";
+    "de" "de";
+  }
+
   server {
     listen       80;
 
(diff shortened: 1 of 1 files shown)


octycs:
Oof, yeah this looks quite confusing. I'd also be okay with a simpler version that might not be 100% standards compatible, such as German is choosen if the header starts with `de`, else English (I mean you can easily change the language if it doesn't fit)

webclient/nginx.conf
```diff
@@ -14,6 +14,40 @@ http {
...
+    "~*.*de,.*en.*" "de";
+    "~*.*en,.*de.*" "en";
+    "~*.*de-.*en.*" "de";
+    "~*.*en-.*de.*" "en";
+    "~*.*de;.*" "de";
+    "~*.*en;.*" "en";
+    "~*.*de,.*" "de";
+    "~*.*en,.*" "en";
+    "~*.*de.*" "de";
+    "~*.*en.*" "en";
```
CommanderStorm, line 42:
```suggestion
    "~*.*de.*en.*" "de";
    "~*.*en.*de.*" "en";
    "~*.*de.*" "de";
    "~*.*en.*" "en";
```
This would be an alternative that still detects the ordering of en and de as well as thier existance. It may have false positives and edge-cases that are not handled properly.

octycs: How would this handle if there was no language header at all (e.g. a bot request), or if it contains neither `de` nor `en`? Can we make English the default then?

CommanderStorm: This is Case-agnostic byw (`"~*` is agnostic vs `"~` which is not agnostic) I cant think of a case where either one would hurt/ benefit from this beaviour. Do you?

CommanderStorm: > How would this handle if there was no language header at all (e.g. a bot request), or if it contains neither `de` nor `en`? Can we make English the default then?

currently `default "de";`. I dont think that we should change this, since not our whole UI (I am thinking about strings we get from the API) is translated.

octycs: As far as I know the standard is something like `lang-REGION`, so technically something like this could be an issue `fr-DE;en-DE`, but I doubt that this is realistic. But I'd make it case-agnostic.

octycs: > currently `default "de";`. I dont think that we should change this, since not our whole UI (I am thinking about strings we get from the API) is translated.

Ah, sorry, didn't see this. Yes, this is also okay for me. It's only that there is a default.

octycs (approved):

PR #55, opened 2022-03-09
Fixed ngnix routing

CommanderStorm:
refactors how the ngnix config is layed out
Furthermore all requests are routed to the correct file

For 404 urls, the 404-view still looks like on 

- https://roomapi.tum.sexy/building/m 
- or https://roomapi.tum.sexy/bug/m

The differing 404 views should be fixed in another PR imo, but we can put this in this PR if you like

### DIFF
--- a/webclient/src/index.html
+++ b/webclient/src/index.html
@@ -4,7 +4,7 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    
+    <meta name="from-view" content="<!-- @echo view -->">
     <!-- @if view="main" -->
     <!-- For some reason currently only works with Chrome -->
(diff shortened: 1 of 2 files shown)


webclient/src/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    
+    <meta name="from-view" content="<!-- @echo view -->">
```
octycs, line 7: I don't mind putting this there, just wondering whether there is a reason for this.

octycs (approved):

webclient/src/index.html
CommanderStorm: Reasoning being, that if at some point the ngnix conifig acts up we might notice quicker where to look :)
Is not nessesary, but I thought this might be a nice to have

CommanderStorm:
can I merge this @octycs ?

octycs:
Yup

PR #64, opened 2022-04-09
Enabled caching for the `api/search/<str:query>` endpoint

### LINKED ISSUE #19: Server-Request Caching
Evaluate if a caching of server-API-requests is feasible and if it would make the API-Requests faster.

This might be done by using the [cached](https://github.com/jaemk/cached)-crate.
See https://stackoverflow.com/questions/63974092/how-to-cache-or-memoize-data-in-actix-web-route
There is an async implementation available as well for this crate instead of the one mentioned in the Stack Overflow post. We should definitively evaluate if this would make a difference/ be eazier.

CommanderStorm:
- [X] the endpoint caches (rounded) 0ms

resolves #19


CommanderStorm:
Open questions: 
 -[] what is `server/test/cmp-210811.json` used for?
 -[] should we also add caching for the `/api/get/<id>` Endpoint? It does not seem like this is needed, since fetching from the dict seems pretty optimal, right?

octycs (approved):

PR #65, opened 2022-04-09
Minor version bumps

CommanderStorm:
@octycs is there a reason for rusttls being used in production?


octycs:
I'm not sure why exactly I included it, but it could be because I previously tried the meilisearch sdk package, which I think required it, but finally didn't work with actix's threading.

PR #67, opened 2022-04-14
mediaTUM

CommanderStorm:
This PR adds the best images from mediaTUM to NavigaTUM.

There are still some issues, that have to be resolved, before this can be merged:

- [x] I'd like your feedback on the values, I chose for `THUMBNAIL_SIZE`
and `HEADER_MAX_SIZE`
- [x] There seem to remain some issues with the image-matching
<details>

```
-- 50 Add image information
Warning: No source information for image '0509.EG.980_0.webp', it will not be used
Warning: No source information for image '0509.EG.980_1.webp', it will not be used
Warning: No source information for image '0509.EG.980_2.webp', it will not be used
Warning: No source information for image '2319.EG.001_0.webp', it will not be used
Warning: No source information for image '5401.01.101K_0.webp', it will not be used
Warning: No source information for image '9001.EG.012_0.webp', it will not be used
Warning: No source information for image '5532.Z1.003_0.webp', it will not be used
Warning: There are images for 'tb-stammgelaende', but it was not found in the provided data, ignoring
Warning: There are images for 'tb-chemie', but it was not found in the provided data, ignoring
Warning: There are images for 'ottobrunn', but it was not found in the provided data, ignoring
```

</details>

- [x] I have not implemented a way of controll how a thumbnail is cropped. See `5304_0.webp` (the mensa) for an example, why this is nessesary. 
My strategy of using the center square of the image works pretty much on every other image, but here it fails.


CommanderStorm:
@octycs what do you think about this way of generating the `thumb` and `header-small` folders?
Arguably it is not free (it adds about 45s to data deployments)

octycs:
> @octycs what do you think about this way of generating the `thumb` and `header-small` folders? Arguably it is not free (it adds about 45s to data deployments)

Could we make the `large` images a bit smaller? From the few I looked at, they had quite a high resolution, and even if we add some kind of image gallery for entries, I think they could (and should) be a lot smaller. Maybe a maximum width/height of 1600px

CommanderStorm:
Yes, we could scale them down. 
I wouldnt go as far as 1600. If we are on a desktop, I would expect a large image to load..

How about we resize to the following:
- thumb (256x256)
- sm (max. 1024px)
- md (max. 1920px) 
- lg (max. 4k) 

and include the images via an [srcset](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-srcset)?

octycs:
Then we should best add some simple local checksum, so that for local dev builds they're only scaled once (This could also be part of another PR).

For the scaling, I think this is okay. But I'd leave some `header` images that have a fixed aspect ratio (512x210px at the moment, could be different) for the info cards, so image loading doesn't cause layout shifts, and to leave them consistent in layout (i.e. no portrait/vertical images there).

CommanderStorm:
Okay, so that I missunderstand your message: Do you think my proposal is good, or what do you mean?
Would you prefer for a layout like this:
- thumb (256x256)
- header (512x210px)
- sm (max. 1024px)
- md (max. 1920px)
- lg (max. 4k)

octycs:
Yes, I think that would be good.

CommanderStorm:
Todo:

- [ ] go through the header images and add offsets, where images look "weird"
- [ ] migrate the webclient to the slightly changed dataformat
- [ ] Document the changes

CommanderStorm:
@octycs I am done with the change.
Now only a code-review and we can ship this

octycs:
Nice, thank you. Yes, I think we can merge this now.

PR #72, opened 2022-04-15
migrated the api to utilise GitHub for feedback-collection

### LINKED ISSUE #17: Implement feedback form to work with GitHub
The current feedback Form only works with GitLab. It would be ideal, if we could design a system, that fits into our privacy statement.
In deployment, the Feedback will be disabled until this issue is resolved.

CommanderStorm:
TODO:
- [x] It is possibly smart to display the github issue link in a small popup when an issue is summitted.
- [x] document these Endpoint in the API Docs
- [x] go through a thorough code review. This is the first time, where I have actually written my own code (instead of heavily reliing on docs/ your code). There are probably some "creative" design choices.

resolves #17

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -2,7 +2,7 @@
 name: Feature request
 about: Suggest an idea for this project
 title: "[Feature]"
-labels: enhancement
+labels: feature
 assignees: ''
 
 ---
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -12,6 +12,7 @@ path = "src/main.rs"
 strip = true
 
 [dependencies]
+octocrab = "0.15.4"
 actix-web = "4.0.1"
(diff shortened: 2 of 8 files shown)


octycs:
I just looked through it and can't see any direct issues. As I saw, you also tested it already. In my opinion we can merge this, unless you have any specific part you're unsure about.

octycs (approved):

CommanderStorm:
I thought you might have a better (more rustacian) some comments about `capitalised_category`, the way I get `github_token`, or some borrow checking stuff.

octycs:
It could be that there are some more elegant ways for stuff like getting the GitHub token, but honestly I'm neither more experienced in rust to tell. It appears to be safe and understandable code to me at least. The possibility to set the GitHub token from the env is certainly a good option.

CommanderStorm:
> The possibility to set the GitHub token from the env is certainly a good option.

That is a requirement, as I can onlly set secure secrets as enviroment variables and not as command attributes ;)

PR #76, opened 2022-04-19
migrated the app to use sqlite as a file-based k/v store

### LINKED ISSUE #74: [Feature] Memory Demand of the server
Currently We need about 430M of RAM (profiled using heaptrack, total ram Usage is about 530M) to hold the JSON_DATA map. This is kind of an problem, since my machine only has 4G of RAM.

Honestly I have no idea. All the solution I have tried either dont borrow or typecheck and I have no idea how to fix that.
One more promissing aproach was using a BTreeMap, but the compiler kept complaining about named livetimes and I could not get him to shut up. Can you (@octycs ) figure out, how to not make the compiler complain? My current state is:
```rust
lazy_static! {
    static ref JSON_DATA: BTreeMap<&'static str, &'static str> = {
        let data = fs::read_to_string("data/api_data.json")
            .expect("Cannot open data file. (not found at 'data/api_data.json')");
        let json_data: serde_json::map::Map<String, serde_json::Value> =
            serde_json::from_str(&data).expect("Could not parse JSON file");
        let mut data_map = BTreeMap::new();
        for (key, value) in json_data.iter() {
            data_map.insert(key.borrow(), value.as_str().unwrap());
        }
        return data_map;
    };
}

#[get("/api/get/{id}")]
async fn get_handler(params: web::Path<String>) -> Result<HttpResponse> {
    let p = params.into_inner();
    let id = p.as_str();
    let data = JSON_DATA.get(id);
    match data {
        Some(data) => Ok(HttpResponse::Ok().json(data)),
        None => Ok(HttpResponse::NotFound().body("Not found")),
    }
}
``` 

Using files to offload the RAM-Requirement

**Additional context**
The real reason, why this is nessesary is, that during deployments (because we do rolling deployments and #27 ) memory sometimes spikes, resulting in OOM events. An OOM Event causes our service to go down for about a minute.

CommanderStorm:
The reasoning is, that the performance is imo sufficient and I cannot get a purely rust based solution to work. See #74 for my struggles.
resolves #74

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -12,11 +12,11 @@ path = "src/main.rs"
 strip = true
 
 [dependencies]
+rusqlite = "0.27.0"
 octocrab = "0.15.4"
 actix-web = "4.0.1"
 awc = "3.0.0"
 cached = "0.34.0"
-lazy_static = "1.4.0"
 serde = { version = "1.0.136", features = ["derive"] }
 serde_json = "1.0.79"
 actix-cors = "0.6.1"
--- a/server/Dockerfile.server
+++ b/server/Dockerfile.server
@@ -1,3 +1,12 @@
+FROM python:3.10.4-bullseye as build-stage
+# get api_data.json
+ADD https://nav.tum.sexy/cdn/api_data.json data/api_data.json
+# For local testing if roomapi is not avalible:
+# follow the data-docs to get api_data.json, copy it to the directory server/data and enable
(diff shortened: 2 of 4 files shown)


octycs:
Seems good to me, and I also think is a better, more future proof solution to #74. We could consider adding some caching there some time maybe (e.g. the `root` element is probably being requested quite often), but only if it turns out to be a bottleneck.

PR #77, opened 2022-04-19
Minor rust tlr

CommanderStorm:
I have changed some minor things, but I would like the rust changes through you.

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "navigatum-server"
 version = "0.1.0"
-authors = ["Markus A <ge75sig@mytum.de>"]
+authors = ["Markus A <ge75sig@mytum.de>", "Frank Elsinga <frank@elsinga.de>"]
 edition = "2021"
 
 [[bin]]
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -1,11 +1,11 @@
 [package]
-name = "server"
+name = "navigatum-server"
 version = "0.1.0"
 authors = ["Markus A <ge75sig@mytum.de>"]
 edition = "2021"
 
 [[bin]]
-name = "server"
+name = "navigatum-server"
 path = "src/main.rs"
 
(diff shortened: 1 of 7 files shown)


server/src/search.rs
```diff
@@ -40,6 +41,7 @@ struct SearchFilter {
 #[derive(Debug)]
 struct SearchInput {
     tokens: Vec<SearchToken>,
+    #[allow(dead_code)]
     filter: SearchFilter,
```
octycs, line 45: Ah yes, these two properties are still unused. I already implemented parsing filters in the search query some time ago (eg. `type:room` or `@garching`), but they're not yet used.

"Regular splits" mean that a search token has been split by a space or similar. Unregular splits are e.g. done on `MW2001` -> `MW 2001` because else Meilisearch has issues. I thought that this field might help when filteing/sorting the results from Meilisearch, but this is also not implemented yet.

octycs (approved):
I added a comment to explain those two unused properties, but just for clarification. You can by the way add you as an author to the `Cargo.toml` if you want.

server/Cargo.toml
```diff
@@ -1,17 +1,18 @@
 [package]
-name = "server"
+name = "navigatum-server"
 version = "0.1.0"
 authors = ["Markus A <ge75sig@mytum.de>"]
```
CommanderStorm, line 4:
```suggestion
authors = ["Markus A <ge75sig@mytum.de>", "Frank Elsinga <frank@elsinga.de>"]
```

CommanderStorm:
I dont see a comment (if you mean a sourcecode comment)
If you mean the github comment I see that one

octycs:
> I dont see a comment (if you mean a sourcecode comment) If you mean the github comment I see that one

Yes, I meant the GitHub comment

CommanderStorm:
Could you document what has to get done for these features to work in an issue instead?
I think, that would be more future-proof :)
Anyway I am merging this PR..

PR #79, opened 2022-04-22
OpenAPI Docs

CommanderStorm:
This PR replaces our documentation via openapi docs.
This PR also fixes some errors in the server that came to light while documenting it.

Todo:
- [x] fix error, that keeps tripping up the server
- [x] validate, that the shema I have written down is actually correct
     - [x] search
     - [x] get
     - [x] get_token
     - [x] feedback
     - [x] healthcheck
- [x] proofread the docs

@octycs I have one problem with the server. This is why I am publishing this PR now, instead of later when I would be more shure, that the `openapi.yaml` I have written is 100% correct.
You can reproduce the issue by 

0. run the server in a different tab
1. installing [schemathesis](https://github.com/schemathesis/schemathesis)
2. `st run --endpoint /api/search/.+ ./openapi.yaml --base-url http://localhost:8080` 

I am not getting anywhere debugging this (the error seems to pop out of nothingness). Could you look into this


CommanderStorm:
Note, that a documentation is online at https://editor.swagger.io/?url=https://raw.githubusercontent.com/TUM-Dev/navigatum/openapi/openapi.yaml

CommanderStorm:
A minimal Testcase is 
```http
GET /api/search/1

**Comment by octycs** [comment]
These errors seem to occur because we don't have a maximum value for `limit`. Schemathesis does requests with limits >30000, so searches for e.g. `1` return so many results that the response exceeds 2MiB, which is the default maximum for `body()` when receiving the response, causing an [`Err(Overflow)`](https://docs.rs/awc/latest/awc/error/enum.PayloadError.html#variant.Overflow).

We can increase this limit to e.g. 10MiB with:
```rust
        .send_json(&post_data)
        .await
        .unwrap()
        .body()
        .limit(10 * 1024 * 1024)
        .await;
```
([Doc](https://docs.rs/awc/latest/awc/struct.ResponseBody.html#method.limit))

But I think it makes more sense to define reasonable maximum values for limit (anything from 100 to 1000, what do you think?)

server/src/search.rs
```diff
@@ -127,24 +135,33 @@ struct MSFacetDistribution {
...
 
-    return do_geoentry_search(&parsed.tokens, args).await;
+fn sanitise_args(args: SearchQueryArgs) -> SanitisedSearchQueryArgs {
+    let max_size = u16::MAX as usize;
+    SanitisedSearchQueryArgs {
+        limit_buildings: min(args.limit_buildings.unwrap_or(5), max_size),
+        limit_rooms: min(args.limit_rooms.unwrap_or(5), max_size),
+        limit_all: min(args.limit_all.unwrap_or(20), max_size),
+    }
 }
```
CommanderStorm, line 165: honestly, both options are fine by me. 
But sanatising user input is always prefeable.
 
Limiting to 1k seems to be a reasonable maximum, that nobody should hit.
`let max_size = u16::MAX as usize` was a bit much in retrospect.

README.md
```diff
@@ -65,6 +72,26 @@ If you want to only run the webclient locally, you can skip the "Data" and
...
+
+We format our api via [openapi-format](https://www.npmjs.com/package/openapi-format).
+```bash
+npm install openapi-format
+openapi-format ./openapi.yaml --output ./openapi.yaml
+```
+
+For validating, that the specification is being followed use the [Swagger Editor](https://editor.swagger.io/?url=https://raw.githubusercontent.com/TUM-Dev/navigatum/main/openapi.yaml) in tandem with  [stoplight](stoplight.io), as they are both very imperfect tools.
+
+To making sure, that this specification is up-to-date and without holes, we run [schemathesis](https://github.com/schemathesis/schemathesis) using the following command on API Server provided by the "Server" step or the public API:
```
octycs, line 85:
```suggestion
To make sure that this specification is up-to-date and without holes, we run [schemathesis](https://github.com/schemathesis/schemathesis) using the following command on API Server provided by the "Server" step or the public API:
```

README.md
```diff
@@ -65,6 +72,26 @@ If you want to only run the webclient locally, you can skip the "Data" and
...
 
+### API
+
+We format our api via [openapi-format](https://www.npmjs.com/package/openapi-format).
+```bash
+npm install openapi-format
+openapi-format ./openapi.yaml --output ./openapi.yaml
+```
+
+For validating, that the specification is being followed use the [Swagger Editor](https://editor.swagger.io/?url=https://raw.githubusercontent.com/TUM-Dev/navigatum/main/openapi.yaml) in tandem with  [stoplight](stoplight.io), as they are both very imperfect tools.
```
octycs, line 83:
```suggestion
To validate that the specification is being followed, use the [Swagger Editor](https://editor.swagger.io/?url=https://raw.githubusercontent.com/TUM-Dev/navigatum/main/openapi.yaml) in tandem with [stoplight](stoplight.io), as they are both very imperfect tools.
```
Not sure about grammar but sound slightly better to me.

server/src/search.rs
```diff
@@ -127,24 +135,37 @@ struct MSFacetDistribution {
...
-    })
+    }
 }
 
 // size=100 seems to be about 10M
 #[cached(size = 500)]
 async fn execute_search(q: String, args: SearchQueryArgs) -> Vec<SearchResultsSection> {
     let parsed = parse_input_query(q);
+    let sanatised_args = sanitise_args(args);
+    return do_geoentry_search(&parsed.tokens, sanatised_args).await;
```
octycs, line 155:
```suggestion
    let sanitised_args = sanitise_args(args);
    return do_geoentry_search(&parsed.tokens, sanitised_args).await;
```

server/src/search.rs
```diff
@@ -127,24 +135,37 @@ struct MSFacetDistribution {
...
+fn clamp_limit(val: usize, min_val: usize) -> usize {
+    // Clamp a value to min_val..1000
+    max(min_val, min(1_000, val))
+}
+
+fn sanitise_args(args: SearchQueryArgs) -> SanitisedSearchQueryArgs {
+    SanitisedSearchQueryArgs {
+        limit_buildings: clamp_limit(args.limit_buildings.unwrap_or(5), 0),
+        limit_rooms: clamp_limit(args.limit_rooms.unwrap_or(5), 0),
+        limit_all: clamp_limit(args.limit_all.unwrap_or(20), 1),
```
octycs, line 167:
```suggestion

fn sanitise_args(args: SearchQueryArgs) -> SanitisedSearchQueryArgs {
    SanitisedSearchQueryArgs {
        limit_buildings: args.limit_buildings.unwrap_or(5).clamp(0, 1_000),
        limit_rooms: args.limit_rooms.unwrap_or(5).clamp(0, 1_000),
        limit_all: args.limit_all.unwrap_or(20).clamp(1, 1_000),
```
We can use Rust's built-in clamp here.

server/src/search.rs
```diff
@@ -1,5 +1,6 @@
 use cached::proc_macro::cached;
 use futures::join;
+use std::cmp::{max, min};
```
octycs, line 3:
```suggestion
```
The compiler complains this is not being used anymore.

server/src/search.rs
```diff
@@ -331,67 +352,48 @@ async fn do_geoentry_search(
...
     let q_default = build_query_string(&search_tokens);
 
-    let res_merged = do_meilisearch(
+    let fut_res_merged = do_meilisearch(
         client.clone(),
         MSSearchArgs {
             q: &q_default,
             filter: None,
-            limit: args.limit_all.unwrap_or(20) as u8, // This is the MeiliSearch default
+            limit: args.limit_all, // This is the MeiliSearch default
```
octycs, line 360:
```suggestion
            limit: args.limit_all,
```

octycs (approved):
Only some small code changes, but seems fine to me. I'll look through the `openapi.yaml` in a moment.

openapi.yaml
```diff
@@ -0,0 +1,1366 @@
...
+    name: GPL v3
+    url: 'https://www.gnu.org/licenses/'
+  version: 1.0.0
+servers:
+  - url: 'https://nav.tum.sexy'
+    description: https
+  - url: 'http://nav.tum.sexy'
+    description: http
+paths:
+  '/api/search/{search_querry}':
```
octycs, line 34:
```suggestion
  '/api/search/{search_query}':
```

openapi.yaml
```diff
@@ -0,0 +1,1366 @@
...
+      operationId: search
+      summary: Search entries
+      description: |
+        This endpoint is designed to support search-as-you-type results.
+
+        Instead of simply returning a list, the search results are returned in a way to provide a richer experience by splitting them up into sections. You might not necessarily need to implement all types of sections, or all sections features (if you just want to show a list). The order of sections is a suggested order to display them, but you may change this as you like.
+
+        Some fields support highlighting the query terms and it uses DC3 (\x19 or \u{0019}) and DC1 (\x17 or \u{0017}) to mark the beginning/end of a highlighted sequence (See [Wikipedia](https://en.wikipedia.org/wiki/C0_and_C1_control_codes#Modified_C0_control_code_sets)). Some text-renderers will ignore them, but in case you do not want to use them, you might want to remove them from the responses (there is no query parameter for this yet).
+      parameters:
+        - name: search_querry
```
octycs, line 45:
```suggestion
        - name: search_query
```

server/src/search.rs
CommanderStorm: Ah, nice to know xD

openapi.yaml
```diff
@@ -0,0 +1,1366 @@
...
+                            name: 5416.EG.010 (IT/Audio)
+                            subtext: 'Interimshörsäle II, Jürgen Manchot-Hörsaalgebäude'
+                            subtext_bold: 010@5416
+                        nb_hits: 23
+                    time_ms: 23
+        '400':
+          description: Invalid Request
+          content: {}
+        '404':
+          description: '`search_querry` is empty. Since searching for nothing is nonsensical, we dont support this.'
```
octycs, line 210:
```suggestion
          description: '`search_query` is empty. Since searching for nothing is nonsensical, we dont support this.'
```

octycs (commented):
I started looking into this, but noticed there is still some modifications needed to the search limits. The current behavior is a bit more complex / the limits don't do exactly what they appear to do.

For example if `limit_rooms` and `limit_buildings` are both 5, but a search returns only rooms, it will return 10 rooms (this is for UX, because the available space of the 5 buildings can be taken up by the 5 rooms).

I'll try to optimize this for public API use.

openapi.yaml
```diff
@@ -0,0 +1,1366 @@
...
+            This could be because we are experiencing technical difficulties or intentional. Please try again later.
+          content: {}
+      tags:
+        - feedback
+  '/cdn/{size}/{id}_{counter}.webp':
+    get:
+      operationId: img_cdn
+      summary: Get title images
+      description: |
+        This endpoint is designed to fetch the images, that are described by the `/api/get/{id}`-endpoint
```
octycs, line 630:
```suggestion
        This endpoint is designed to fetch the images, that are described by the `/api/get/{id}`-endpoint.
```

openapi.yaml
```diff
@@ -0,0 +1,1366 @@
...
+          description: 'The uri you are trying to request is unreasonably long. neither ids, nor any other parameter has more than 30 chars..'
+          content: {}
+      tags:
+        - cdn
+  '/cdn/maps/{source}/{format}/{id}.webp':
+    get:
+      operationId: maps_cdn
+      summary: Get title images
+      description: |
+        This endpoint is designed to fetch the images, that are described by the `/api/get/{id}`-endpoint
```
octycs, line 716:
```suggestion
        This endpoint is designed to fetch the images, that are described by the `/api/get/{id}`-endpoint.
```

octycs (approved):
Just two periods missing. Apart from this this is done imo.

CommanderStorm:
One last thing to discuss in this PR:

One thing that I have not touched, is
- https://github.com/TUM-Dev/navigatum/blob/openapi/data/data_format_event.yaml
- and https://github.com/TUM-Dev/navigatum/blob/openapi/data/data_format_geo-entry.yaml

Should we keep those? The openapi does kind of duplicate those. Also there are fields in openapi.yaml that are not in here.
I am divided on this:
- leaving it woud be good, since they are denser.
- removing it would mean, that documentaton and actual response would be closer together.

CommanderStorm:
Before this can be merged the changes from #84 have to be rectified. This is why I rebased and squashed some of my commits now

octycs:
> One last thing to discuss in this PR:
> 
> One thing that I have not touched, is
> 
>     * https://github.com/TUM-Dev/navigatum/blob/openapi/data/data_format_event.yaml
> 
>     * and https://github.com/TUM-Dev/navigatum/blob/openapi/data/data_format_geo-entry.yaml
> 
> 
> Should we keep those?

The `data_format_event.yaml` can be discarded for now, there isn't even much in there (I started drafting event functionality for e.g. lectures or student council events).

Though I would prefer to keep the `data_formt_geo-entry.yaml` for now since it is (in my opinion) in a format that allows to quickly get a look at the api data. It also still has some information that is missing in that detail in the `openapi.yaml` as far as I could see, especially the explanations of the entry types (`joined_building` etc.).

CommanderStorm:
Sorry for the duplicate review. I had to fix some of the things that cropped up.
There is one issue left, this seems to be a raw-data problem or am I misunderstanding the signals I added? 
image

is minimum: 0 a bad requirement for x/y?

octycs:
> There is one issue left, this seems to be a raw-data problem or am I misunderstanding the signals I added?

Oh, that issue is new to me, I didn't ever check that. That's an issue of the automatic Roomfinder map assignment, that doesn't take rotation into account (which is with very few exceptions small, so it seemed to not be a problem). But I'm wondering why e.g. `1:500, FMI Garching BT08 1.OG` is even assigned for the MI, because the automatic assignment requires all children to be within all maps for a building.

So yes, 0 as a minimum makes sense, the issue is the automatic assignment, which needs to be revised.

octycs (approved):
Apart from this assignment issue the changes appear fine to me.

PR #83, opened 2022-04-23
Cacheing for image `resize_and_crop`

CommanderStorm:
This PR adds the Functionality, to cache image `resize_and_crop` operations. 
The Cache-key is a sha256 of the fileconent and a sha256 the offsets.


octycs (approved):
I've added two small fixes that I found on my local tests, but else it works fine. If you approve, we can merge this imo.

PR #90, opened 2022-04-26
Maps Sortation and deduplication

### LINKED ISSUE #37: [Bug] duplicate maps
**Describe the bug**
Some maps are included twice.
image
an example for this is
https://github.com/TUM-Dev/navigatum/blob/main/data/external/maps/roomfinder/171.gif
and https://github.com/TUM-Dev/navigatum/blob/main/data/external/maps/roomfinder/170.gif

**To Reproduce**
1. Go to `https://roomapi.tum.sexy/site/mri`
2. Click on 'Lagepläne'
4. See multiple `1:500000, Münchner Süden` selections, which look virtually identical

**Expected behavior**
No duplicate Maps

**Screenshots**
image

CommanderStorm:
This PR upgrades the maps by adding an uniqueness check and a merger for offending maps. 
It also sorts the maps by scale and area

resolves #37

### DIFF
--- a/data/processors/maps.py
+++ b/data/processors/maps.py
@@ -237,7 +237,10 @@ def _extract_available_maps(entry, custom_maps, maps_list):
 
 
 def _merge_str(s1: str, s2: str):
-    """Merges two strings. The Result is of the format common_prefix s1/s2 common_suffix"""
+    """
+    Merges two strings. The Result is of the format common_prefix s1/s2 common_suffix.
+    Example: "Thierschbau 5. OG" and "Thierschbau 6. OG" -> "Thierschbau 5/6. OG"
+    """
     if s1 == s2:
         return s1
     prefix = os.path.commonprefix((s1, s2))
--- a/data/processors/maps.py
+++ b/data/processors/maps.py
@@ -1,6 +1,7 @@
 import json
 import math
 import copy
+from typing import Any
 
 import utm
 import yaml
@@ -42,8 +43,8 @@ def assign_coordinates(data):
                 error = True
                 continue
             if "utm" in entry["coords"] \
-               and (   entry["coords"]["utm"]["easting"]  == 0.
-                    or entry["coords"]["utm"]["northing"] == 0.):
+                    and (entry["coords"]["utm"]["easting"] == 0.
+                         or entry["coords"]["utm"]["northing"] == 0.):
                 print(f"Error: UTM coordinate is zero for '{_id}': "
                       f"{entry['coords']}")
                 error = True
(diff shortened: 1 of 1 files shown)


data/processors/maps.py
```diff
@@ -124,112 +130,187 @@ def assign_roomfinder_maps(data):
...
+    for m in maps_list:
+        # Assuming coordinates in Central Europe
+        if m["latlonbox"]["south"] < lat_coord < m["latlonbox"]["north"] and \
+                m["latlonbox"]["west"] < lon_coord < m["latlonbox"]["east"]:
+            available_maps.append(m)
+    return available_maps
+
+
+def _merge_str(s1: str, s2: str):
+    """Merges two strings. The Result is of the format common_prefix s1/s2 common_suffix"""
```
octycs, line 240:
```suggestion
    """
    Merges two strings. The Result is of the format common_prefix s1/s2 common_suffix.
    Example: "Thierschbau 5. OG" and "Thierschbau 6. OG" -> "Thierschbau 5/6. OG"
    """
```
This is quite fancy, but wasn't directly clear to me.

data/processors/maps.py
```diff
@@ -124,112 +130,187 @@ def assign_roomfinder_maps(data):
...
-    
+
     for _id, entry in data.items():
         if entry["type"] == "root":
             continue
-        
-        if len(entry.get("maps", {}).get("roomfinder", {}).get("available", [])) > 0:
+
+        if entry.get("maps", {}).get("roomfinder", {}).get("available", []):
             continue
```
octycs, line 143: (see review comment)

octycs (approved):
This is definitely good, and we could merge this I think, but it has probably only a small effect as I've noticed when looking for example at `0503.07.316`.
This room has one of the name-merged maps, but the merge has no effect in practice because of the `if` I marked below. It already has its Roomfinder maps assigned, so the code here is skipped.

If we didn't skip here though, the auto assign includes all floors, because we don't directly have floor information for Roomfinder maps:
image

So this PR is fine I think, but another issue is the proper assignment of maps taking care of the floors. One possible solution could be to automatically determine this by looking at the rooms in the original Roomfinder assignments. If they are all on the same floor, this is probably a map for this floor.

data/processors/maps.py
```diff
@@ -124,112 +130,187 @@ def assign_roomfinder_maps(data):
...
-                entry.setdefault("maps", {})["roomfinder"] = roomfinder_map_data
-            else:
-                # TODO: 5510.02.001
-                roomfinder_map_data = entry.setdefault("maps", {}).get("roomfinder", {})
-                roomfinder_map_data.update(building_parent["maps"]["roomfinder"])
-                roomfinder_map_data["is_only_building"] = True
 
+        # Use maps from parent building, if there is no precise coordinate known
+        entry_accuracy_building = entry["coords"].get("accuracy", None) == "building"
+        if entry_accuracy_building and entry["type"] in {"room", "virtual_room"}:
```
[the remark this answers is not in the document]
CommanderStorm: Yes, this is just for readability. I did not like, how this was a two-line if statement and how semi-cryptic the statement originally looked..

CommanderStorm:
> This is definitely good, and we could merge this I think, but it has probably only a small effect as I've noticed when looking for example at `0503.07.316`. This room has one of the name-merged maps, but the merge has no effect in practice because of the `if` I marked below. It already has its Roomfinder maps assigned, so the code here is skipped.
> 
> If we didn't skip here though, the auto assign includes all floors, because we don't directly have floor information for Roomfinder maps: image
> 
> So this PR is fine I think, but another issue is the proper assignment of maps taking care of the floors. One possible solution could be to automatically determine this by looking at the rooms in the original Roomfinder assignments. If they are all on the same floor, this is probably a map for this floor.

could you write an issue for this?
You probably have more context to write a better issue, since you wrote the data portion 😉

PR #98, opened 2022-05-05
Generate sitemap

### LINKED ISSUE #91: [Feature] Enable google indexing
Google currently fails to parse our website.

To remidy this, we should

- add a sitemap
      - [ ] the sitemap should look at the currently deployed site to figure out if the content was updated
      - [ ] for this to be effective, we should consider adding the image hashes to the api
      - [ ] It should include an index, that links to all pages generated by the webclient and the data urls in a different file.
       The max-sice of these maps has to be respected.
- [ ] add canonical urls 
- [ ] fix why google does not find our subsites

Preferably we should implement all three.
Just fixing issue bulletpoint 3 would probably be sufficient, but since google is currently prefering navigatum.tum.sexy instead of nav.tum.sexy at least 2 should be adressed as well.

**Additional context**
The issue google has could be because of a too strict robots.txt
Any disallowances before have been lifted in https://github.com/TUM-Dev/navigatum/commit/d4aea9541e6d8d4c1e54ce3d0a60c6bf4fd9a7db

octycs:
Resolves #91.

The sitemap generation doesn't take into account changes in the source images, however it does detect updates if the image attribution or source URL has changed, or if an image has been added etc. So I think this shouldn't be a problem in practice.

### DIFF
--- a/data/processors/sitemap.py
+++ b/data/processors/sitemap.py
@@ -24,7 +24,8 @@ def generate_sitemap():
     # so it's unlikely that we'll hit this limit without adding a lot of
     # data. But for the case that a new type of entry is introduced, the
     # sitemap is split into one for rooms and one for the rest.
-    # Note that the root element is included, which will link to the main page.
+    # Note that the root element is not included, because it just redirects
+    # to the main page.
     sitemaps = {
         "room": [],
         "other": [],
@@ -37,6 +38,9 @@ def generate_sitemap():
 
     changed_count = 0
     for _id, entry in new_data.items():
+        if entry["type"] == "root":
+            continue
+
         sitemap_name = entry["type"] if entry["type"] in sitemaps else "other"
 
         url = f"https://nav.tum.sexy/view/{_id}"
@@ -54,11 +58,8 @@ def generate_sitemap():
         # rooms in general in the range 0 to 9, so we just add 1, divide by 10
         # and clamp to 1.0 for rooms.
         # For buildings etc. that are always >= 10, we just subtract 5
-        # to get some kind of relative measure. Also for root (which is
-        # downrated in search) we set 1.0 by hand.
-        if entry["type"] == "root":
-            priority = 1.0
-        elif entry["type"] == "room":
+        # to get some kind of relative measure.
+        if entry["type"] == "room":
             priority = min((entry["ranking_factors"]["rank_combined"] + 1) / 10, 1.0)
         else:
             priority = min((entry["ranking_factors"]["rank_combined"] - 5) / 10, 1.0)
--- a/data/processors/sitemap.py
+++ b/data/processors/sitemap.py
@@ -43,7 +43,19 @@ def generate_sitemap():
 
         sitemap_name = entry["type"] if entry["type"] in sitemaps else "other"
 
-        url = f"https://nav.tum.sexy/view/{_id}"
+        # Just copied from the webclient. The webclient doesn't care about
+        # the prefix – if it is wrong it'll be corrected (without a redirect).
+        # However this way search engines can already index the final URL.
+        url_type_name = {
+            "campus": "campus",
+            "site": "site",
+            "area": "site",
+            "building": "building",
+            "joined_building": "building",
+            "room": "room",
+            "virtual_room": "room"
+        }[entry["type"]]
+        url = f"https://nav.tum.sexy/{url_type_name}/{_id}"
 
         if _id not in old_data or entry != old_data[_id]:
             last_changed = datetime.now()
--- a/data/processors/sitemap.py
+++ b/data/processors/sitemap.py
@@ -93,9 +93,7 @@ def _download_online_sitemaps(sitemap_names):
                     lastmod = child.find(f"{xmlns}lastmod")
                     if loc is not None and lastmod is not None:
                         sitemaps[name][loc.text] = datetime.fromisoformat(lastmod.text)
-        # We gracefully catch any problems (e.g. a 404 HTTPError), because
-        # having the old sitemaps is not critical.
-        except Exception as e:
+        except urllib.request.HTTPError as e:
             print(f"Warning: Failed to download sitemap '{name}': {e}")
 
     return sitemaps
@@ -127,7 +125,7 @@ def _write_sitemapindex_xml(fname, sitemaps):
         loc = ET.SubElement(sitemap_el, "loc")
         loc.text = f"https://nav.tum.sexy/sitemap-data-{name}.xml"
         changefreq = ET.SubElement(sitemap_el, "changefreq")
-        changefreq.text = "daily"  # data sitemaps might change frequently
+        changefreq.text = "weekly"
 
     # Because sitemaps cannot be hierarchical, we have to include the
     # webclient sitemap here as well.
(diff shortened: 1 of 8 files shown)


octycs:
The nginx config is not yet tested, but that might be easiest when deployed.

webclient/gulpfile.js
```diff
@@ -455,6 +456,27 @@ function copy_assets() {
...
+                       path.dirname = 'about';
+                       path.extname = '';
+                   } else {
+                       path.dirname = '';
+                   }
+               }))
+               .pipe(sitemap({
+                   siteUrl: 'https://nav.tum.sexy/',
+                   fileName: 'sitemap-webclient.xml',
+                   changefreq: 'weekly',
```
CommanderStorm, line 473: More than the data, we dont change our about pages weekly. If we change something major, we can rerequest an indexing by google.
I would suggest `monthly`

webclient/nginx.conf
```diff
@@ -78,10 +78,17 @@ http {
...
+    location = /sitemap-webclient.xml { access_log off; }  # webclient sitemap
     location = /favicon.ico { root /app/assets/;}
     # These Files are intenitonally not supported
     location = /adds.txt { log_not_found off; access_log off; }
     location = /app-ads.txt { log_not_found off; access_log off; }
+    
+    # sitemaps hosted on the cdn (webclient sitemap already catched above)
+    location ~ ^/sitemap {
+      access_log off;
+      proxy_pass https://nav.tum.sexy/cdn$request_uri;
```
CommanderStorm, line 90: Wouldnt it be more efficient to directly link to the cdn? Why do we have to add this extra latency?

data/processors/sitemap.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def _write_sitemapindex_xml(fname, sitemaps):
+    """ Write the sitemapindex XML """
+    sitemapindex = ET.Element("sitemapindex")
+    sitemapindex.set("xmlns", "http://www.sitemaps.org/schemas/sitemap/0.9")
+    for name in sitemaps.keys():
+        sitemap_el = ET.SubElement(sitemapindex, "sitemap")
+        loc = ET.SubElement(sitemap_el, "loc")
+        loc.text = f"https://nav.tum.sexy/sitemap-data-{name}.xml"
```
CommanderStorm, line 128: as stated below; Why dont we link to the cdn directly? We shoudl add a comment if there is some reason..

data/processors/sitemap.py
```diff
@@ -0,0 +1,142 @@
...
+                root = defusedET.fromstring(sitemap_str)
+                for child in root.iter(f"{xmlns}url"):
+                    loc = child.find(f"{xmlns}loc")
+                    lastmod = child.find(f"{xmlns}lastmod")
+                    if loc is not None and lastmod is not None:
+                        sitemaps[name][loc.text] = datetime.fromisoformat(lastmod.text)
+        # We gracefully catch any problems (e.g. a 404 HTTPError), because
+        # having the old sitemaps is not critical.
+        except Exception as e:
+            print(f"Failed to download sitemap '{name}': {e}")
```
CommanderStorm, line 99:
```suggestion
            print(f"Warning: Failed to download sitemap '{name}': {e}")
```

data/processors/sitemap.py
```diff
@@ -0,0 +1,142 @@
...
+        lastmod = ET.SubElement(url, "lastmod")
+        lastmod.text = sitemap_entry["last_changed"].isoformat()
+        priority = ET.SubElement(url, "priority")
+        priority.text = str(round(sitemap_entry["priority"], 2))
+
+    root = ET.ElementTree(urlset)
+    root.write(fname, encoding="utf-8", xml_declaration=True)
+
+
+def _write_sitemapindex_xml(fname, sitemaps):
```
CommanderStorm, line 121: fname is a constant and function is only used once, so there is no real need to pass it. 
It could be named in the docstring.. 
I dont have a strong opinion on this. Current variable passing has upsides as well.

CommanderStorm (commented):
I have left some inline comments. This is mostly good to go, just some minor things that need changing.

CommanderStorm:
Could you also add canonical urls as mentioned in #91?
For example our canonical url for olimpiapark is https://nav.tum.sexy/campus/olympiapark and not https://nav.tum.sexy/view/olympiapark

This tripps google up for urls like https://navigatum.tum.sexy

webclient/nginx.conf
octycs: My main concern was search engine bots not liking that redirect, because extra latency is not an issue for sitemaps. Though a redirect feels cleaner.

CommanderStorm: No, I did not say a redirect, 
I said directly linking to them :)
Google does not require that the sitemaps are all under the root URL, or do they?

data/processors/sitemap.py
```diff
@@ -0,0 +1,142 @@
...
+    """ Write the sitemapindex XML """
+    sitemapindex = ET.Element("sitemapindex")
+    sitemapindex.set("xmlns", "http://www.sitemaps.org/schemas/sitemap/0.9")
+    for name in sitemaps.keys():
+        sitemap_el = ET.SubElement(sitemapindex, "sitemap")
+        loc = ET.SubElement(sitemap_el, "loc")
+        loc.text = f"https://nav.tum.sexy/sitemap-data-{name}.xml"
+        changefreq = ET.SubElement(sitemap_el, "changefreq")
+        changefreq.text = "daily"  # data sitemaps might change frequently
```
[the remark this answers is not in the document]
octycs: I was thinking the same, probably doesn't really matter much.

data/processors/sitemap.py
octycs: It could be constant, though I preferred that the filename strings are all at the same location this way.

webclient/nginx.conf
octycs: Let's just try

data/processors/sitemap.py
```diff
@@ -0,0 +1,142 @@
...
+
+        url = f"https://nav.tum.sexy/view/{_id}"
+
+        if _id not in old_data or entry != old_data[_id]:
+            last_changed = datetime.now()
+            changed_count += 1
+        else:
+            # Try to look up the last changed date in the old sitemap
+            last_changed = old_sitemaps.get(sitemap_name, {})\
+                                       .get(url, datetime.now())
```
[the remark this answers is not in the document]
octycs: The only case when `old_sitemaps[sitemap_name]` does not exist is, when the sitemap is new, and thus there is not old sitemap for that name. That applies to this PR for example, because currently there is no sitemap online.

The second point is correct though, it should increment the counter.

CommanderStorm:
also we should add our **main sitemap** here: https://github.com/TUM-Dev/navigatum/blob/sitemap/webclient/src/.well-known/robots.txt

Syntax:
```
Sitemap: https://nav.tum.sexy/cdn/sitemap.xml

File: webclient/gulpfile.js [hunk 8]
```diff
@@ -455,6 +456,27 @@ function copy_assets() {
...
+                   lastmod: false,
```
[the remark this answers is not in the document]
octycs: I wasn't sure whether last modified dates will be set correctly inside the Docker build and thought it doesn't really matter anyway with a low change frequency. However we can just try it out.

octycs:
> Could you also add canonical urls as mentioned in #91?

The issue with navigatum.tum.sexy should be solved with the sitemap (each URL in the sitemap is [considered canonical](https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls#sitemap-method)). And for the `/view` vs `campus` etc. that should be solved in https://github.com/TUM-Dev/navigatum/pull/98/commits/2e453d95570eb3097230b138dba591a9f7490554.

CommanderStorm (approved):
Lets see, if it works

PR #108, opened 2022-05-09
Image-showcase

### LINKED ISSUE #89: [Feature] image carousel
We currently supply multiple images for some buildings. In the frontend we however show only one size (`header`)

It would be cool to have a image carousel pop up, once a user presses on the image with all of our images.
The image should be marked up via [googles image licence structured data](https://developers.google.com/search/docs/advanced/structured-data/image-license-metadata).
It would also be preferable to include the images via a sourceset, since we include multiple sizes for each image (`sm`,`md`,`lg`)

Keep it as is and only show one image. Not ideal, but it works

CommanderStorm:
this PR adds an image showcase, for all the beautiful images we have.

UI samples:
Screenshot_20220509_233042
Screenshot_20220509_233008

resolves #89

### DIFF
--- a/webclient/src/views/view/i18n-view.yaml
+++ b/webclient/src/views/view/i18n-view.yaml
@@ -50,3 +50,14 @@ view_view:
     no_floor_overlay:
       de: "Für den angezeigten Raum gibt es leider keine Indoor Karte."
       en: "There is unfortunately no indoor map for the displayed room."
+  slideshow:
+    header:
+      de: "Bilder-Showcase"
+      en: "Image Showcase"
+    close: {de: "Schließen", en: "Close"}
+    image_alt:
+      de: "Bild, das das Gebäude zeigt"
+      en: "Image showing the building"
+    author: {de: "Autor", en: "Author"}
+    source: {de: "Quelle", en: "Source"}
+    license: {de: "Lizenz", en: "License"}
--- a/webclient/src/views/view/i18n-view.yaml
+++ b/webclient/src/views/view/i18n-view.yaml
@@ -56,7 +56,7 @@ view_view:
       en: "Image Showcase"
     close: {de: "Schließen", en: "Close"}
     image_alt:
-      de: "Bild, das das Gebäude zeigt"
+      de: "Ein Bild welches das Gebäude zeigt"
       en: "Image showing the building"
     author: {de: "Autor", en: "Author"}
     source: {de: "Quelle", en: "Source"}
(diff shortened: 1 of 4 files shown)


octycs:
I'm having some issues with the CSS. Below the images there is the license / source text, but cropped:
image

However the visible source section below that doesn't update when I change the image.

CommanderStorm:
> I'm having some issues with the CSS. Below the images there is the license / source text, but cropped: image
> 
> However the visible source section below that doesn't update when I change the image.

Good catch. My browser did not display that.

The fix for the sidebar issue however is less pretty:
I rely on the undocumented (!) `modal-fullheight`
from https://github.com/picturepan2/spectre/blob/master/src/_modals.scss

CommanderStorm:
I think all the issues should be fixed now.
Could you recheck if it works on your end?

The CSS-issue of the images loading in/out weird (if aspact ratio is not a match) is nothing that I can fix as far as I can tell, but this could also be me doing frontend stuff...

octycs:
That text issue is solved now. But I'm still only seeing the source of the first image when sliding through them.

That CSS issue is a bit annoying, yeah. But I also couldn't find a quick fix. It's hardcoded into spectre. The only we could do easily is turn the animation there off.

octycs:
> But I'm still only seeing the source of the first image when sliding through them.

Ah okay, that issue is only there when I use the buttons at the bottom. When I'm using the right/left arrows it works

CommanderStorm:
Where (browser, url) is this the case?
Every big building with different sources (not all images have distinct sources) works on my machine.

octycs:
E.g. `/building/mw` with Firefox 100.0 and Chromium (Version 101.0.4951.54 (Official Build) Manjaro Linux (64-bit))

octycs:
I'll try to look into it tomorrow

CommanderStorm:
Sorry Per André for the accidental ping...

octycs:
> https://github.com/TUM-Dev/navigatum/pull/108/commits/dd0823e68fae5d586214b78b5e9c4df105a77275

Yup that solves it.

octycs (approved):
I think it's fine like this, though we might turn animations off or find a fix for that sometimes glitchy look.

CommanderStorm:
> I think it's fine like this, though we might turn animations off or find a fix for that sometimes glitchy look.

Sadly If we want to fix it we will have to do that ourselves, since spectre.css is not anylonger maintained.

This can however be a main-commit-bugfix or another PR. I am merging this and creating an issue

PR #110, opened 2022-05-10
Improved Polyfill loading

CommanderStorm:
This PR is going to be a controversial one with you I assume.
I dropped support for IE.

I would also like to discuss https://caniuse.com/?search=fetch and if we really need a polyfill for this. Concretely: Do we support Opera Mini?

One caveat I noticed: `babel_targets` does not affect what polyfills are packaged. Do you know why this is the case?

resolves #110

### DIFF
--- a/webclient/gulpfile.js
+++ b/webclient/gulpfile.js
@@ -114,7 +114,7 @@ gulp.task('main_css', gulp.series(compile_main_scss, compile_spectre_scss/*, mer
 
 // --- Main JS Pipeline ---
 function build_app_core_js() {
-    return gulp.src(['src/core.js'])
+    return gulp.src(['src/core.js', 'src/detect-webp.js'])
                .pipe(concat('app-core.js'))
                .pipe(preprocess({
                     context: {
--- a/webclient/gulpfile.js
+++ b/webclient/gulpfile.js
@@ -63,8 +63,7 @@ var babel_targets = {
     "browsers": [
         "last 2 versions",
         "not dead",
-        "> 0.2%",
-        "IE>=10"
+        "> 0.2%"
     ]
 };
 
--- a/webclient/gulpfile.js
+++ b/webclient/gulpfile.js
@@ -399,22 +399,23 @@ function extract_polyfills() {
 }
 
 function insert_polyfills() {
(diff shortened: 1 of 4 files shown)


octycs:
I'm less concerned of IE, because (as you once said) no-one should be using it anymore. But I'm more concerned because of old smartphones with old browsers, which appear (to what I see in the physics department) more common.

In any case, I think we should add some error in case people are using an outdated browser that this is the reason why they are seing an infinite loading spinner.

What I'd like was if we some kind of declare what browsers and versions we want to support, so we can keep track of it and evaluate for each feature that may break (i.e. the last 5 years or specific Mac/iOS Safari releases / Android API levels. Manually installed Firefox or Chrome are usually not the issue ...).

octycs:
I have only limited data from other websites on what browsers people use and no data how it's like at TUM, but it appears some people do use Opera. Samsung Internet is a browser we should support.

image

octycs:
I did two changes, that I hope optimizes this for modern browsers:
1. Polyfills are now loaded only in browsers that do not support modules. That includes all browsers that don't support fetch.
2. I moved the WebP detection to the core JS sources, because I think this is the only extra polyfill that we do need to detect differently. This slows down page views for old Safaris, but speeds it up for modern browsers, because they don't need to download polyfills.js for this anyway.

CommanderStorm:
Looks way better.
This should improve things a lot.
I have access to an old iMac this evening and I am going to test, wether this polyfill indeed works as I would expect.

There is still the issue, that babel does not really seem to be concearned with `babel_targets`. Do you know if this is intentional/ a bug?
Because of what you did, I am not overly concearned with loading this extra code as I origianlly was. Opera Mini is esentially the only browser this affects, so we really dont need to spend any more time on this than nessesary.

octycs:
I don't really know why Babel doesn't seem consider `babel_targets`. Actually, I was just quite happy to get any polyfills running back then that worked, so I didn't test it. I think this is (besides the pending Safari test) ready to be merged.

octycs (approved):

CommanderStorm:
It works on my roommates iMac (as expected, but still).
Merging it 🚀

Issue #111, opened 2022-05-11
[Feature] CI


CommanderStorm:
**Is your feature request related to a problem? Please describe.**
We should add a CI-Pipeline.

**Describe the solution you'd like**
This Pipeline should have 2 paralel steps, with paralel subtasks:
- execute precommit:
  - pylint
  - mypy
  - black
  - eslint
  - stylint
  - cargo fmt
  - ... 
- server:
  - cargo clippy
  - cargo test

**Describe alternatives you've considered**
including cargo clippy/cargo test in the precommit skript, but this is suboptimal since they take ages to finish.

**Additional context**
The CI should not block CD. Sounds counterintuitive, but at least until the bugs we noticed in #104 are fixed, the CI would fail => no deployments.

PR #112, opened 2022-05-13
Linting

CommanderStorm:
This PR adds Code-Linting.

For this PR to get merged, this work still has to be done:
- [x] gain a consensus, what linters we do want to include
- [x] gain a consensus, how they should be configured
- [x] let the linter run (and fix the stuff the linter finds)

The current configuration has a few downsides:
- it adds two files, that need to be at the project root. This may be consolidated in the future into one, but currently this is impossible
- There are many preferences in here… You will find some of them questionable, but this is something we should discuss
- currently this is only a linting suite. 
  A test suite will be added later in another PR.
  Doing both in a CI will be another PR.

Running the CI:
- `python -m venv venv`
- `source venv/bin/activate`
- `pip install -r requirements_dev.txt`
- `pre-commit run -a`

a step in the path towards #111


.pre-commit-config.yaml
```diff
@@ -0,0 +1,104 @@
...
+      - id: check-ast
+      - id: check-toml
+      - id: check-yaml
+        args:
+          - --allow-multiple-documents
+        exclude: "data/data_format_geo-entry.yaml"
+      - id: check-json
+      - id: check-executables-have-shebangs
+      - id: check-shebang-scripts-are-executable
+      - id: check-merge-conflict
```
octycs, line 14: Do we need this, or doesn't GitHub do this already?

CommanderStorm: I dont know if they do. I also could not find a documentation on this..
We can remove this, if you'd like.. 

This should never trigger, as with modern IDEs merging stuff is not a huge issue anymore imo..

This is something that is quite obvious in PRs and I assume, that neither you nor me will forget this, while merging a PR..
So this may be essentially wasted CPU-time..

webclient/src/core.js
```diff
@@ -184,61 +180,63 @@ var navigatum = (function () {
...
+            });
         },
         getBaseData: function(id) {
-            return this.getData(id, false)
+            return this.getData(id, false);
         },
         getExtendedData: function(id) {
-            return this.getData(id, true)
+            return this.getData(id, true);
         },
```
CommanderStorm: I removed extended data here, because it is unused

openapi.yaml
```diff
@@ -1079,7 +1079,7 @@ components:
         - height
         - width
         - x
-        - y
+        - 'y'
```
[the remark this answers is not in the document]
CommanderStorm: I have no idea.. it really likes it this way..

CommanderStorm:
> In general I can approve, but the main thing that bothers me is that sometimes I find the compact, space-adjusted formatting more readable than the version that adheres to the maximum line length.
 
In general, I like having a normalized way of having this data.
Otherwise, these huge JSON files just become blobs (which are unreviewable), I cannot reason about…
For some of the YAML files, I agree: they should probably be reorganized or ignored.

> I think this is personal preference, but what do you think about excluding some of the files from line length linting (if that is possible)?

Definitely. Some files are already excluded from this (`openapi.yaml`)

openapi.yaml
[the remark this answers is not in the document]
octycs: Hm, okay. It's weird but it doesn't bother me.

PR #116, opened 2022-05-14
Add hash to JS files and assets for caching

octycs:
This renames JS files in release builds e.g. from `app-core-for-view-main-de.min.js` to `cache_f52f42ee.app-core-for-view-main-de.min.js` based on the hash of the file (we might change that filename format, it's just starting with `cache_` so it looks nicer in the nginx config).

This way we can set long cache times for them without worrying about changes.

### DIFF
--- a/webclient/gulpfile.js
+++ b/webclient/gulpfile.js
@@ -17,6 +17,7 @@ const markdown    = require('gulp-markdown');
 const preprocess  = require('gulp-preprocess');
 const purgecss    = require('gulp-purgecss');
 const rename      = require('gulp-rename');
+const revAll      = require("gulp-rev-all");
 const sass        = require('gulp-sass')(require('node-sass'));
 const sitemap     = require('gulp-sitemap');
 const splitFiles  = require('gulp-split-files');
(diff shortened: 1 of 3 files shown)


webclient/gulpfile.js
```diff
@@ -456,6 +457,26 @@ function copy_assets() {
...
+                       includeFilesInManifest: [".js"],
+                       dontRenameFile: [".html"],
+                       transformFilename: function (file, hash) {
+                           var ext = path.extname(file.path);
+                           return "cache_" + hash.substr(0, 8) + "." + path.basename(file.path, ext) + ext;
+                       },
+                   }))
+                   .pipe(gulp.dest('build'))
+    else
+        return done();
```
CommanderStorm, line 476: This is somewhat personal preference, but I like to omit else-cases if possible
```suggestion
    if (config.target !== "release")
        return done();
    return gulp.src(['build/index-*.html', 'build/js/*.js'])
               .pipe(revAll.revision({
                   // Currently .js only, because important css is inlined, and postloaded
                   // css is deferred using preload, which revAll currently doesn't detect
                   includeFilesInManifest: [".js"],
                   dontRenameFile: [".html"],
                   transformFilename: function (file, hash) {
                       var ext = path.extname(file.path);
                       return "cache_" + hash.substr(0, 8) + "." + path.basename(file.path, ext) + ext;
                   },
               }))
               .pipe(gulp.dest('build'))
```

webclient/gulpfile.js
```diff
@@ -456,6 +457,26 @@ function copy_assets() {
 }
 gulp.task('assets', copy_assets);
 
+// --- Revisioning Pipeline ---
+function revision_assets(done) {
+    if (config.target === "release")
+        return gulp.src(['build/index-*.html', 'build/js/*.js'])
+                   .pipe(revAll.revision({
+                       // Currently .js only, because important css is inlined, and postloaded
+                       // css is deferred using preload, which revAll currently doesn't detect
```
CommanderStorm: What about the content of `/assets/` i.e. `.webp`, `.svg`, `.png`, `.ico`?
Could we include the here too?
This could simplify the cache-configuration a bit in the ngnix file

CommanderStorm (approved):

webclient/gulpfile.js
octycs: I think we could also add them to caching, too. I'll try

octycs:
I've added them too. Is this how you would change the nginx config?

webclient/nginx.conf
```diff
@@ -123,13 +123,12 @@ http {
       try_files /index-view-view-$THEME-$LANG.html /404.html;
     }
 
-    location ^~ /assets/ {
-      expires 1d;
+    location ~ ^/(js|assets)/cache_.*$ {
+      expires 30d;
```
CommanderStorm, line 127: We may want to increase this value in the future, but for now this seems fine

CommanderStorm (approved):

CommanderStorm:
I think this is ready to be merged.

webclient/nginx.conf
octycs: Nginx has the option `expires max` iirc, but yeah, we can change that in the future.

PR #118, opened 2022-05-16
re-enabled a limited form of of coodinate check

CommanderStorm:


data/processors/maps.py
```diff
@@ -126,6 +126,23 @@ def assign_coordinates(data):
...
+def check_coords(input_data):
+    """ Check for issues with coordinates """
+
+    for iid, data in input_data.items():
+        if data["type"] == "root":
+            continue
+
+        if data["coords"]["lat"] == 0. or data["coords"]["lon"] == 0.:
+            print(f"{iid}: lat and/or lon coordinate is zero. If unknown, omit this field.")
+            continue
```
octycs, line 138:
```suggestion
            print(f"{iid}: lat and/or lon coordinate is zero. Please provide an accurate coordinate!")
            exit(1)
```
I think it would be better to fail at this point, because we shouldn't have wrong data compiled in (I know this is old code from me)

data/processors/maps.py
```diff
@@ -126,6 +126,23 @@ def assign_coordinates(data):
...
+
+        if "utm" in data["coords"] and (
+                data["coords"]["utm"]["easting"] == 0. or
+                data["coords"]["utm"]["northing"] == 0.):
+            print(f"{iid}: utm coordinat is zero. If unknown, omit this field.")
```
octycs, line 143:
```suggestion
            print(f"{iid}: utm coordinate is zero. There is very likely an error in the source data (UTM coordinates are either from the Roomfinder or automatically calculated).")
            exit(1)
```

octycs (approved):
I've changed the behavior to fail in case there is a zero coordinate, because this is certainly wrong. What do you think?

CommanderStorm:
Agree, but I would like this to be an Error.

octycs (approved):
Yes, a RuntimeError is certainly better.

PR #120, opened 2022-05-17
Coordinate picker

### LINKED ISSUE #107: [Feature] Coordinate Picker
It would be nice for users to quickly give feedback on where a room is really located.

I would like a coordinate picker in the feedback section, where users can interactively input the coodinates of a room

Having users do that manually, but this is unrealisitc.

**Additional context**
The current workflow is kind of sucky..

CommanderStorm:
This PR implements a coodinate Picker.
resolves #107 

Todo:
- [x] add translations
- [x] figgure out a way of creating a draggable custom marker (I have tried the simple ways, something is not working; I have to spend some quality time debuging this)
- [x] add a better styling than this:
     image
image
image

since you are more into design: what UI would you (@octycs ) prefer?


CommanderStorm:
The reason why a custom logo is not draggable is, because we use images in a div container instead of (svg) images for this usecase. I think this should be another PR

CommanderStorm:
Call to action:
image

webclient/src/i18n.yaml
```diff
@@ -55,6 +55,30 @@ feedback:
...
+      de: "Generelles Feedback über diese website"
+      en: "General Feedback about this website"
+    bug:
+      de: "Welchen Fehler hast du gefunden? Wo hast du ihn gefunden? Bitte gib eine genaue Beschreibung an"
+      en: "Which bug did you find? Where did you find it? Please provide a detailed description"
+    features:
+      de: "Features, die du gerne auf dieser Website haben würdest"
+      en: "Features, you would like to see on this website"
+    search:
+      de: "Feedback zur Suche. Was war dein Suchbegriff? Was hättest du erwartet, das das Ergebins ist?"
```
octycs, line 69:
```suggestion
      de: "Feedback zur Suche. Was war dein Suchbegriff? Was hättest du als Ergebnis erwartet?"
```

webclient/src/i18n.yaml
```diff
@@ -55,6 +55,30 @@ feedback:
...
+      en: "Feedback about the search. What was your search query? What did you expect to see?"
+    entry:
+      de: "Feedback zu einem Eintrag. Wir können Räume/Gebäude/Standorte hinzufügen und alle Daten die du siehst (Namen, Coordinaten, Adressen, ...) anpassen. Was hast können wir verbessern?"
```
octycs, line 72:
```suggestion
      de: "Feedback zu einem Eintrag. Wir können Räume/Gebäude/Standorte hinzufügen und alle Daten, die du siehst (Namen, Coordinaten, Adressen, ...) anpassen. Was können wir verbessern?"
```

webclient/src/views/view/view-view.js
```diff
@@ -190,6 +192,30 @@ navigatum.registerView('view', {
...
+            this.map.interactive.marker2.remove();
+            this.map.interactive.marker2 = null;
+            let actionMsg="${{_.feedback.coordinatepicker.add_coordinate}}$"
+            if (this.view_data.coords.accuracy !== "building")
+                actionMsg="${{_.feedback.coordinatepicker.correct_coordinate}}$"
+            const body=`${actionMsg}:\n` +
+                "\`\`\`\n" +
+                `"${this.view_data.id}": {coords: {lat: ${location.lat}, lon: ${location.lng}}},\n`+
+                "\`\`\`";
+            open_feedback("entry", "[" + this.view_data.id + "]: ", body);
```
octycs, line 217: I think we could prefill the subject as well.

octycs (commented):
Just some typos at the moment. I will test styling (esp. on mobile) later.

One thing though (we talked about it): Should be add a note to collect multiple edits in bulk? Else someone might hit the feedback rate limit.

Some possible idea: When selecting a coordinate, it is added into local storage, so that opening the dialog just reads all coordinates from local storage (the part in `view-view.js`). Then we can say on the feedback form, that your edit is temporarily saved and you can also close the dialog and add it for another room. Only when hitting submit (for type 'entry'), the local storage is cleared, or after a timeout (e.g. 12h).

CommanderStorm:
> One thing though (we talked about it): Should we add a note to collect multiple edits in bulk? Else someone might hit the feedback rate limit.
> 
> Some possible idea: When selecting a coordinate, it is added into local storage, so that opening the dialog just reads all coordinates from local storage (the part in `view-view.js`). Then we can say on the feedback form, that your edit is temporarily saved and you can also close the dialog and add it for another room. Only when hitting submit (for type 'entry'), the local storage is cleared, or after a timeout (e.g. 12h).

I will have a look into it this comment this evening, if this would be a large change, I would like for this to be its own PR, but maybe this is just a small change..

CommanderStorm:
I added the workflow as requested. please have a look at https://github.com/TUM-Dev/navigatum/pull/120/commits/2c8b90a02e8edecc3471851854ac094d45709652
Is this something we want, or should this be reverted?

Pro:
- The amount of tokens requested per client should be reduced, if the user is using multiple tabs

Con:
-  if the user submits in one tab, this will not affect the token in the other tab => when the user tries to submit this token the user sees an error

CommanderStorm:
I also do not know, if the toasts I added are to verbose.. I am somewhat divided on this

octycs:
> Is this something we want, or should this be reverted?

Oh right, I didn't think about using multiple tabs. In my opinion it's better yes, because it should (or am I wrong?) only be a problem if you had the feedback form already open in Tab B when clicking Submit in Tab A, and it should be easily fixed by copying the text and reopening the feedback form.

octycs:
> The current draft seems a bit overwhealming but I dont know of a shorter way to phrase the information

I'm not entirely sure where that last sentence about buildings etc. comes from. As far as I understand, the body is always just the list of collected edits, so it would only be a title/text mismatch (add vs. correct coordinate), but no data lost. Or am I missing something?

CommanderStorm:
> Or am I missing something?

What I am imaginging is:

1. User edits one coordinate and corrects the name
2. closes the Feedback form
3. opens it on another page (custom things the user did to the body are not present anymore)

octycs:
What about a warning:
> If you want to correct several coordinates at the same time, you can close this window and edit more coordinates. <br> We will save your changes locally for 12 hours. <br> Warning, that's only valid for coordinates!

> Falls du mehrere Koordinaten gleichzeitig korrigieren willst, kannst du dieses Fenster schließen und weitere Koordinaten editieren. Wir speichern deine Veränderungen für 12 Stunden bei dir lokal. <br> Achtung, das gilt nur für Koordinaten!

Is this clear enough?

octycs (approved):

PR #129, opened 2022-05-24
Fixed map assignment

### LINKED ISSUE #97: [Bug] Some maps are assigned by mistake
**Describe the bug**
As noticed in https://github.com/TUM-Dev/navigatum/pull/79#issuecomment-1117950972 we sometimes assign maps, which make not a lot of sense
That's an issue of the automatic Roomfinder map assignment, that doesn't take rotation into account (which is with very few exceptions small, so it seemed to not be a problem).

**To Reproduce**
Run the the data compilation pipeline

**Expected behavior**
x of assinged entries should Never be negative

CommanderStorm:
- makes the maps processor more reasonable in terms of technical dept
- fixes the assignment issue from #97 

resolves #97

### DIFF
--- a/data/compile.py
+++ b/data/compile.py
@@ -3,7 +3,7 @@ import os
 import json
 
 from utils import convert_to_webp
-from processors import areatree, images, maps,coords, merge, patch, roomfinder, search, sections, sitemap, structure, tumonline
+from processors import areatree, images, maps, coords, merge, patch, roomfinder, search, sections, sitemap, structure, tumonline
 
 DEBUG_MODE = "GIT_COMMIT_SHA" not in os.environ.keys()
 
--- a/data/compile.py
+++ b/data/compile.py
@@ -73,7 +73,8 @@ def main():
 
     logging.info("-- 45 Roomfinder maps")
     maps.assign_roomfinder_maps(data)
-    maps.check_roomfinder_maps_default(data)
+    maps.remove_non_covering_maps(data)
+    maps.assign_default_roomfinder_map(data)
     maps.build_roomfinder_maps(data)
 
     logging.info("-- 46 Overlay maps")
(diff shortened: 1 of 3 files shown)


CommanderStorm:
Note that there is a trade-off in here:
I did not like how complex the assignment and building of maps was, so I cut it up into smaller subtasks.
This removes the ability to detect obviously wrong coordinates like for example https://nav.tum.sexy/room/0104.U1.421 
For this I would suggest that we should open an issue and solve this in a different, more explicite sainity check for the coordinates

data/processors/maps.py
```diff
@@ -14,133 +12,6 @@
...
-            continue
-
-        if "coords" in entry:
-            # While not observed so far, coordinate values of zero are typical for missing
-            # data so we check this here.
-            if entry["coords"].get("lat", None) == 0. or entry["coords"].get("lon", None) == 0.:
-                logging.error(f"Lat and/or lon coordinate is zero for '{_id}': {entry['coords']}")
-                error = True
-                continue
-            if "utm" in entry["coords"] \
```
CommanderStorm, line 50: I removed this check, since covered in another module

data/compile.py
```diff
@@ -3,7 +3,7 @@
 import json
 
 from utils import convert_to_webp
-from processors import areatree, images, maps, merge, patch, roomfinder, search, sections, sitemap, structure, tumonline
+from processors import areatree, images, maps,coords, merge, patch, roomfinder, search, sections, sitemap, structure, tumonline
```
octycs, line 6:
```suggestion
from processors import areatree, images, maps, coords, merge, patch, roomfinder, search, sections, sitemap, structure, tumonline
```

octycs (approved):

PR #135, opened 2022-05-27
Debug overlay

CommanderStorm:
This PR adds a "debug overlay" feature for better debugging of roomfinder maps.
This works by adding a button in the fronend, which if clicked switches the Overlay maps to one made by the roomfidner team.

I think, that this belongs in the roomfinder, since it should greatly simplify the debugging of roomfinder-maps issues.

There are currently a few bugs left to fix:
image
The real problem is, that I do not know, what is meant by the rotation field. Since you have experiment with that: What rotation is meant by this?

I am also not super happy with how the button is styled, and I would like to have these maps more transparent to make the difference more apparent.


CommanderStorm:
I am giving up on this, as I cannot make calculating border coordinates from bboxes work

Issue #87, opened 2022-04-24
[Feature] Add links to OSM / Google Maps


octycs:
**Is your feature request related to a problem? Please describe.**
Except copying the address, there is currently no way to use an external service to navigate to the displayed building / site.
(See also #86)

**Describe the solution you'd like**
Add a small link to OSM / Google Maps on the Info Card, based on the entries coordinates.

**Describe alternatives you've considered**
/

**Additional context**
Tasks could include:
- Add relevant HTML/CSS in `webclient/src/views/view/view-view.inc|.scss`. The data is already there, and every entry is guaranteed to have coordinate data.

CommanderStorm:
I think this can be generalised:
A share button would probably be nice.
This should include a way to toggle if you share a google maps/osm/navigatum link.

Note that there is already some link copy code for this feature:
![image](https://user-images.githubusercontent.com/26258709/169603287-63d2a7a2-ee3f-44c1-890c-0d36c1cc66cd.png)

octycs:
Closed by #136

PR #136, opened 2022-06-07
Add external links popup to OSM / Google Maps

octycs:
This adds a small popup with links to OSM and Google Maps. The Google Maps links _should_ directly open the app on Android and according to the internet also on iOS.

Currently the popup is implemented CSS only. This appeared to work, we have to see whether this causes issues.

image

Solves #87

### DIFF
--- a/webclient/src/views/view/view-view.inc
+++ b/webclient/src/views/view/view-view.inc
@@ -72,16 +72,26 @@
       </div>
       <div class="column col-auto col-ml-auto">
         <button class="btn btn-link btn-action btn-sm"
-                title="${{ _.view_view.header.external_link }}$" >
+                title="${{ _.view_view.header.external_link.tooltip }}$" >
           <svg xmlns="http://www.w3.org/2000/svg" width="14" height="14" viewBox="0 0 3.704 3.704" fill="none" stroke="#0065bd" stroke-width=".529" stroke-linecap="round">
             <path d="M2.912 2.179v1.26H.267V.794h1.197" stroke-linejoin="round"/><path d="M1.407 2.297l2.03-2.03"/><path d="M2.352.268h1.085v1.085" stroke-linejoin="round"/>
           </svg>
         </button>
         <div class="link-popover">
+          <strong>${{ _.view_view.header.external_link.open_in }}$</strong>
           <a class="btn" target="_blank"
              :href="'https://www.openstreetmap.org/?mlat=' + view_data.coords.lat + '&mlon=' + view_data.coords.lon + '#map=17/' + view_data.coords.lat + '/' + view_data.coords.lon + '&layers=T'">OpenStreetMap</a><br>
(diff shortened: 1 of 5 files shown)


webclient/src/views/view/view-view.inc
```diff
@@ -71,6 +71,18 @@
...
+                title="${{ _.view_view.header.external_link }}$" >
+          <svg xmlns="http://www.w3.org/2000/svg" width="14" height="14" viewBox="0 0 3.704 3.704" fill="none" stroke="#0065bd" stroke-width=".529" stroke-linecap="round">
+            <path d="M2.912 2.179v1.26H.267V.794h1.197" stroke-linejoin="round"/><path d="M1.407 2.297l2.03-2.03"/><path d="M2.352.268h1.085v1.085" stroke-linejoin="round"/>
+          </svg>
+        </button>
+        <div class="link-popover">
+          <a class="btn" target="_blank"
+             :href="'https://www.openstreetmap.org/?mlat=' + view_data.coords.lat + '&mlon=' + view_data.coords.lon + '#map=17/' + view_data.coords.lat + '/' + view_data.coords.lon + '&layers=T'">OpenStreetMap</a><br>
+          <a class="btn" target="_blank"
+             :href="'https://www.google.com/maps/search/?api=1&query=' + view_data.coords.lat + '%2C'+ view_data.coords.lon">Google Maps</a>
```
CommanderStorm, line 84: I would prefer a "Open in XY" (obviously localised)
Reasoning:
We are using a share button and we are not shareing to OSM/Maps, but opening the location in this app.

CommanderStorm (changes requested):
- Would it be possible to add a "Copy Link" button, that literally copies the current website link to the clipboard like the one we already have?

- On mobile may want to support https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share
On some platforms (Namely anything linux) this is however "creatively" implemented.
On Mobile this looks like this:
Screenshot_20220608-001710_Chrome
As a link we might share sometthing like `geo:latitude,longitude` as suggested [here](https://stackoverflow.com/questions/15053004/how-to-generate-a-location-link-that-will-open-in-default-mapping-app) and/or the sites link.

Request changes is because I currently consider the share icon a bit misleading without any sharing functionality... 
This is still an improvement and could be merged, but I would prefer an a bit different messaging.
The second point can go into an issue and live there, as essentially I am not shure how this really should be implemented, but maybe you have a good idea for this dilemma

CommanderStorm:
An example for a less supported platform for the API I referenced can be seen here:
image
(Manjaro, Linux, Chrome)

octycs:
Now this also supports Sharing:
image

The Geo URI is behind the 'other app' button, and the browser share functionality is only visible if the browser supports sharing (which neither Chrome nor Firefox do on my system for some reason).

CommanderStorm (approved):

Issue #140, opened 2022-06-10
[Feature] Bigger Map (Fullscreen-Button)


CommanderStorm:
**Is your feature request related to a problem? Please describe.**
Sometimes it would be nice to have a larger map (especially on desktop).
![image](https://user-images.githubusercontent.com/26258709/173084287-a4ec3315-7e75-4bd6-baf3-73470b5fbb68.png)

**Describe the solution you'd like**
I think, a fullscreen-button would be a great addition.
Only adding such a button on desktop is something we should discuss

**Describe alternatives you've considered**
Making the map on desktop take a bit more room might be a good solution too

**Additional context**

octycs:
Solved by #142

PR #142, opened 2022-06-11
Maximize Map on Desktop

octycs:
This adds the fullscreen button to the map on desktop as well, but instead of going completely fullscreen, the map is only maximized:

image

It was a bit tricky to extend the FullscreenControl to do that (all my attempts of ES5 or ES6 class inheritance failed or ended in a lot of badly readable code, so I decided to modify the FullscreenControl instance).

Note it is intentional, that the map remains maximized when you go somewhere else via search. This feels a bit more like a maps app.

Solves #140

### DIFF
--- a/webclient/src/modules/interactive-map.js
+++ b/webclient/src/modules/interactive-map.js
@@ -5,7 +5,7 @@ navigatum.registerModule("interactive-map", (function() {
 
     // Because mapboxgl might not be loaded yet, we need to postpone
     // the declaration of the FloorControl class
-    function FloorControlInit() {
+    function floorControlInit() {
         // Add Evented functionality from mapboxgl
         FloorControl.prototype = Object.create(mapboxgl.Evented.prototype);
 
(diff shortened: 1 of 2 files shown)


webclient/src/modules/interactive-map.js
```diff
@@ -211,11 +211,39 @@ navigatum.registerModule("interactive-map", (function() {
...
+            // is resized (not relevant for users but for developers).
+            const is_mobile = window.matchMedia &&
+                              window.matchMedia("only screen and (max-width: 480px)").matches;
+
+            fs_control = new mapboxgl.FullscreenControl({
+                container: is_mobile ? document.getElementById("interactive-map")
+                                     : document.getElementById("interactive-map-container")
+            });
+            fs_control.__onClickFullscreen = fs_control._onClickFullscreen;
+            fs_control._onClickFullscreen = function() {
```
CommanderStorm, line 226: What is de distinction between `__onClickFullscreen` and `_onClickFullscreen`?
Is there a better way to name this? Seems an easy source for future errors..

CommanderStorm (approved):
can be merged as is

webclient/src/modules/interactive-map.js
octycs: This is just the "backup" of the mapboxgl default fullscreen handler, that I extended by overwriting this function. I've renamed it to make it clearer.

PR #154, opened 2022-06-15
improved feedback description handling

### LINKED ISSUE #134: [Bug] Submitting feedback via the feedbackform strips newlines
**Describe the bug**
When feedback is submitted via the feedbackform, 

**To Reproduce**
1. Go to the feedback button
2. enter 
image

3. See #132 

**Expected behavior**
The `\n` should be replaced with `\n  ` in the backend, to be more in line what I expect a user to expect

**Additional context**
I opened this as an issue, to get some feedback if my view on this is valid, or if we should stick with the current status quo

CommanderStorm:
added a improved feedback description parsing and feedback description tests

resolves #134

### DIFF
--- a/server/src/feedback.rs
+++ b/server/src/feedback.rs
@@ -250,9 +250,6 @@ mod description_tests {
         for i in 0..10 {
             assert_eq!(clean_feedback_data("abcd", i), "abcd".truncate(i));
         }
-        assert_eq!(clean_feedback_data("abcd", 1), "a");
-        assert_eq!(clean_feedback_data("abcd", 3), "abc");
-        assert_eq!(clean_feedback_data("abcd", 6), "abcd");
     }
     #[test]
     fn special_cases() {
--- a/server/src/feedback.rs
+++ b/server/src/feedback.rs
@@ -248,7 +248,7 @@ mod description_tests {
     #[test]
     fn truncate_len() {
         for i in 0..10 {
-            assert_eq!(clean_feedback_data("abcd", 0), "");
+            assert_eq!(clean_feedback_data("abcd", i), "abcd".truncate(i));
         }
         assert_eq!(clean_feedback_data("abcd", 1), "a");
         assert_eq!(clean_feedback_data("abcd", 3), "abc");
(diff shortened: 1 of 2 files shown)


server/src/feedback.rs
```diff
@@ -213,10 +214,50 @@ fn parse_request(req_data: &Json<FeedbackPostData>) -> (String, String, Vec<Stri
...
+        assert_eq!(clean_feedback_data("a\nb  \nc", 9), "a  \nb  \nc");
+        assert_eq!(clean_feedback_data("a      \nb", 9), "a  \nb");
+        assert_eq!(clean_feedback_data("a\n\nb", 9), "a  \n  \nb");
+        assert_eq!(clean_feedback_data("a\n   b", 9), "a  \n   b");
+    }
+    #[test]
+    fn truncate_len() {
+        for i in 0..10 {
+            assert_eq!(clean_feedback_data("abcd", 0), "");
+        }
```
octycs, line 252: Is there a reason this is in a `for`?

CommanderStorm: Ups.. reason was that I wanted to reduce my code-footprint :)

What I really meant is:
```suggestion
        for i in 0..10 {
            assert_eq!(clean_feedback_data("abcd", i), "abcd".truncate(i));
        }
```

server/src/feedback.rs
```diff
@@ -213,10 +214,50 @@ fn parse_request(req_data: &Json<FeedbackPostData>) -> (String, String, Vec<Stri
...
+        assert_eq!(clean_feedback_data("a\n   b", 9), "a  \n   b");
+    }
+    #[test]
+    fn truncate_len() {
+        for i in 0..10 {
+            assert_eq!(clean_feedback_data("abcd", i), "abcd".truncate(i));
+        }
+        assert_eq!(clean_feedback_data("abcd", 1), "a");
+        assert_eq!(clean_feedback_data("abcd", 3), "abc");
+        assert_eq!(clean_feedback_data("abcd", 6), "abcd");
```
octycs, line 255: Should we leave these then?

CommanderStorm:
```suggestion
```

PR #161, opened 2022-07-05
Tokio changes

CommanderStorm:
- move to tokio::main
- change to a tokio mutex, as it supports await

The second one is the motivation for this PR


server/src/core/search/search_executor/preprocess.rs
```diff
@@ -159,15 +159,16 @@ pub(super) fn tokenize_input_query(q: &str) -> Vec<InputToken> {
 
         if (!within_quotes && c.is_whitespace() && i > token_start) ||  // whitespace
            ((within_quotes || !c.is_whitespace()) && i+c.len_utf8() == q.len()) ||  // end of string
-           (c == '"')  // end of quotes
+           (c == '"')
```
octycs, line 162: It's not important, but it might be better to undo the formatting changes here, because they are also affected by #112. Else we can merge this I think.

octycs (approved):

server/src/core/search/search_executor/preprocess.rs
CommanderStorm: While we could do that, they are just some things `cargo fmt` wants, and not in any way controversial, so I think we should pull them out of #112 and just merge them here. They are not conflicting anyway.

PR #163, opened 2022-07-17
MW ground floor

CommanderStorm:
I have added a map for the MW.
Things missing:
- [ ] figure out a way to `Convert to LatLonQuad` without losing the “Squareness” or rotation.  
image
- [ ] assign coordinates for the rooms
- [ ] Decide, whether this map warrants a non-overlay map.

Neither of these are blockers in and of itself, but we can make them into blockers.


octycs:
> [ ] figure out a way to `Convert to LatLonQuad` without losing the “Squareness” or rotation.

So far I've just adjusted until it looked correct, but after some trial and error I managed to create a working conversion now (only valid for local approximations of course):
```python
import math

n, e, s, w, r = 48.266976, 11.671295, 48.264346, 11.667244, -10

center = ((e+w)/2, (n+s)/2)
theta = math.radians(center[1])
alpha = math.radians(r)

def rotate(lat, lon):
    # Convert to local, rectangular x/y grid using standard mathematical axes directions.
    # We need to theta to compute the local aspect ratio of the lat/lon grid (~2:3 in Munich)
    x = (lon - center[0]) * math.cos(theta)
    y =  lat - center[1]

    return center[0] + (x * math.cos(alpha) - y * math.sin(alpha)) / math.cos(theta), \
           center[1] + x * math.sin(alpha) + y * math.cos(alpha)


nw = rotate(n, w)
ne = rotate(n, e)
se = rotate(s, e)
sw = rotate(s, w)

print(nw[0], nw[1])
print(ne[0], ne[1])
print(se[0], se[1])
print(sw[0], sw[1])
```

This prints the coordinates as our format (and mapbox) requires them in the correct order.

octycs:
But I don't think this needs to be part of this PR, we could merge this right away

octycs (approved):

data/sources/46_overlay-maps.yaml
```diff
@@ -141,10 +141,10 @@
...
     parent: "mw"
     box:
-      - [11.667506, 48.266657]
-      - [11.671514, 48.266191]
-      - [11.671084, 48.264587]
-      - [11.667094, 48.265051]
+      - [11.667629, 48.267174]
+      - [11.671599, 48.266688]
+      - [11.670956, 48.264145]
+      - [11.666920, 48.264598]
```
CommanderStorm, line 147: Corrected the coordinates as by suggestion
```suggestion
      - [11.667617801785777, 48.26719015728]
      - [11.67160725799323, 48.26672188711042]
      - [11.670921198214224, 48.264131842720005]
      - [11.66693174200677, 48.26460011288958]
```

CommanderStorm:
> * [ ]  Decide, whether this map warrants a non-overlay map.

@octycs As a follow up: What do you think about this?

octycs:
I think having a non-overlay map would be good as well (or at least I don't know why we shouldn't want one)

PR #166, opened 2022-07-19
Python reformattings

CommanderStorm:
These are the changes to the python files, to make #112 more reviewable

### DIFF
--- a/data/compile.py
+++ b/data/compile.py
@@ -14,7 +14,6 @@ from processors import (
     sitemap,
     structure,
     tumonline,
-    export,
 )
 from utils import convert_to_webp
 
--- a/data/compile.py
+++ b/data/compile.py
@@ -1,9 +1,22 @@
+import json
 import logging
 import os
-import json
 
+from processors import (
+    areatree,
+    coords,
+    images,
+    maps,
+    merge,
+    patch,
+    roomfinder,
+    search,
+    sections,
+    sitemap,
+    structure,
+    tumonline,
+)
 from utils import convert_to_webp
-from processors import areatree, images, maps, coords, merge, patch, roomfinder, search, sections, sitemap, structure, tumonline
 
 DEBUG_MODE = "GIT_COMMIT_SHA" not in os.environ.keys()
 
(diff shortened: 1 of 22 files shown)


data/external/downloader.py
```diff
@@ -578,63 +598,62 @@ def _get_cache_path(fname):
...
+    with open(path, "w", encoding="utf-8") as file:
+        json.dump(data, file)
 
 
 def _get_tumonline_api_url(base_target):
-    return "https://campus.tum.de/tumonline/{}/NC_{}".format(
-        base_target,
-        str(random.randint(0, 9999)).zfill(4),
-    )
+    # TODO: WTF is this ???
```
octycs, line 657: I have absolutely no idea, observed it in the API

data/compile.py
```diff
@@ -64,39 +64,21 @@ def main():
...
-    maps.build_roomfinder_maps(data)
+    maps.roomfinder_maps(data)
 
     logging.info("-- 46 Overlay maps")
     maps.add_overlay_maps(data)
 
-    logging.info(f"-- 50 convert {images.IMAGE_BASE} to webp")
-    convert_to_webp(images.IMAGE_BASE)
-    logging.info("-- 51 resize and crop the images for different resolutions and formats")
+    logging.info("-- 50 resize and crop the images for different resolutions and formats")
```
octycs, line 79:
```suggestion
    logging.info("-- 50 Convert, resize and crop the images for different resolutions and formats")
```
I would add explicitly that this converts the images as well

octycs (approved):
Thank you for splitting #112. I have looked over all changes and cannot see any issues for the Python part. I added two comments, but else I think this is ready for merge.

octycs:
Just to know, is this intended just for reference? Should I comment on #112 instead?

data/external/downloader.py
CommanderStorm:
```suggestion
    # I have no idea, what this magic_string is, or why it exists.. Usage is the same as from TUMOnline..
```

CommanderStorm:
I changed the base to main and am merging just the python reformattings ;)

PR #167, opened 2022-07-19
Js scss reformttings

CommanderStorm:
These are the changes to the js/scss files, to make https://github.com/TUM-Dev/navigatum/pull/112 more reviewable


webclient/src/variables.scss
```diff
@@ -9,26 +11,27 @@ $theme-accent: #0065bd;  // TUM blue
...
-$toast-btn-bg-active: #0004;
+$feedback-overlay-bg: rgba(247, 248, 249, 75%);
+$feedback-box-shadow: 0 .2rem .5rem rgba(48, 55, 66, 30%);
+$toast-btn-bg: rgba(0, 0, 0, 13%);
+$toast-btn-bg-hover: rgba(0, 0, 0, 2%);
+$toast-btn-bg-active: rgba(0, 0, 0, 27%);
 $card-highlighted-bg: #f7f7f7;
 $text-gray: #aaa;
-$border-default: #e1e1e1;
+$border-color: #e1e1e1;
```
octycs, line 28: `border-color` is a spectre.css variable, we shouldn't shadow it here.
```suggestion
$border-color-default: #e1e1e1;
```

webclient/src/variables.scss
```diff
@@ -38,66 +41,55 @@ $code-bg: #fcf2f2;
...
-    $dark-color: #9aa0aa !global;//#3e396b;
+    $dark-color: #9aa0aa !global;// #3e396b;
 
     // Also sets body background
     $light-color: #17181a !global;
 
     //      #0d0e13,         #13141a (also footer), link colors: #59b2ff,
     //    .. 1px solid #e1e1e140 on border
-    $border-color: $border-default !global;
+    $border-color: $border-color !global;
```
octycs, line 91: This statement makes no sense. `$border-color` is a variable of spectre.css, and `$border-default` was our own variable so far. We should at least make it more clear that this is not our own variable and move all `border-color` statements into the "Overwriting" sections

webclient/src/core.js
```diff
@@ -2,464 +2,441 @@
...
+            navigatum.getModule("autocomplete").then((c) => {
+              c.onInput(e.srcElement.value);
             });
-        },
-        getBaseData: function(id) {
-            return this.getData(id, false);
-        },
-        getExtendedData: function(id) {
-            return this.getData(id, true);
-        },
```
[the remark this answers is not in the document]
octycs: Yes, that is a legacy and can be removed. :+1:

octycs (changes requested):
I have tried to look over all files. Sometimes the diff failed to understand the changes and hard to read, so I hope I haven't missed anything that it not reformatting.

There is only one thing that I think we need to keep, that is the `navigatum.cloneState` function instead of `structuredClone`, because Samsung Internet doesn't support it yet. Alternatively we could add a polyfill into core.js so that it is always available (even if polyfills.js is not loaded)

webclient/src/variables.scss
CommanderStorm:
```suggestion
    $border-color: $border-default !global;
```

CommanderStorm:
I noiced that the gulpfile has stopped processing spectre. I will look into this on monday :)

octycs:
> I noiced that the gulpfile has stopped processing spectre. I will look into this on monday :)

Just to be sure, have you done npm update after the change in the package.json :sweat_smile:

octycs:
It was related to the `border-color` variables in sass and should be fixed now.

webclient/src/history-states.js
```diff
@@ -8,106 +8,107 @@
...
-        historyPushState.bind(history)(data, title, url);
-    };
-    addEventListener('popstate', function (e) {
-        // If navigation is history navigation (click on back/forward),
-        // the 'popstate' event is emitted before 'beforeResolve()'.
-        // So in this case, we need to temporarily store the old state index,
-        // so that the saveCurrentViewState() call in beforeResolve() saves the
-        // state to the correct state in the history.
-        history.lastStateIndex = history.stateIndex;
+      window.history.states[stateIndex][0][stateDataSymbol] = structuredClone(
```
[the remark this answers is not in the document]
octycs: I cannot test it, but the polyfill in https://github.com/TUM-Dev/navigatum/pull/167/commits/4615b5a209166b1636e562cb5930b7240a7ad3c1 appears fine

octycs (approved):

PR #175, opened 2022-07-31
Add coordinates for most rooms in the MW ground floor

octycs:
This PR adds the coordinates for most rooms in the MW ground floor (~300 rooms). I have excluded rooms of the following types:
- Schacht
- Stromversorgung
- Elektroverteiler
- Flur
- Installationsraum (Gebäudeleittechnik)
- Server / Serverraum
- Gasflaschenraum

But included:
- Aufzug and Lastenaufzug (possibly relevant for barrier free functionality)
- Treppe / Treppenraum
- Windfang
- Foyer
- Vorraum WC
- Regieraum

Some small parts are still missing because I noticed some errors in our data, in the provided data and in the map.  
- In our data some rooms are missing that exist in TUMOnline, and we need to find out why
Examples:
5510.EG.012A (0012A, Büro)
5510.EG.099t (Treppe Hof 0)
5505.EG.529B (0529B, Büro)
5506.EG.699V (0699V, Treppe)  
  
  --> Moved to #177 

Also there are some errors in our map:
- [x] The building part of [0050](https://nav.tum.sexy/room/5510.EG.050) (EDV-Pool) and others are missing
- [x] The room separators in the area of 0305 and 0306 (building part 3, south-east) are wrong
- [x] 0533b (above 0533) is no room but a floor
- [x] 0403 is only one room (two are shown in the map)
- [x] Building part 8 has room separators missing
- [x] The floor should end before 0234

Finally there are errors in the TUMOnline data:
- 5505.EG.506A (Rechnerraum-Studenten) should end with 506M
- 5505.EG.533A (Bibliothek) should not have an A at the end
- 5504.EG.403J (Bibliothek) has the architect id 0403P, which is two times on the map (need to find out which one is correct)

There are also errors in the source map that I was able to correct.


octycs:
I have created #177 because it appears also rooms in other buildings are missing. The problem with TUMOnline data is that at the moment, it is better to take over the incorrect data from there, because if we change IDs it could make it even harder to find the rooms.

octycs:
I'd merge right away because this PR mainly only adds data

PR #187, opened 2022-08-12
New data

CommanderStorm:
I scraped TUMonline for new data.

This PR is a draft because this URL trips the downloader up:
https://portal.mytum.de/campus/roomfinder/getRoomPlacemark?b_id=2927&mapid=217

Do you know what is the matter with this parameter?

You can reproduce this by moving all except the roomfinder json to the caching folder.

### DIFF
--- a/data/sources/02_rooms-extended.yaml
+++ b/data/sources/02_rooms-extended.yaml
@@ -314,7 +314,9 @@
 "5508.EG.808": { coords: { lat: 48.265906, lon: 11.667831 } }
 "5508.EG.899A": { coords: { lat: 48.265678, lon: 11.667768 } }
 "5508.EG.899B": { coords: { lat: 48.265773, lon: 11.667895 } }
-"5508.02.801": { coords: { lat: 48.265856, lon: 11.667780 } } # Ernst-Schmidt-Hörsaal
+"5508.02.801":  # ID is probably wrong in TUMonline, but we need to use it for compatibility
+  name: "5508.01.801 (Ernst-Schmidt-Hörsaal)"
+  coords: { lat: 48.265856, lon: 11.667780 } }
 
 "5510.EG.001": { coords: { lat: 48.265424, lon: 11.670841 } } # Gustav-Niemann-Hörsaal
 "5510.EG.001M": { coords: { lat: 48.265558, lon: 11.670974 } }
--- a/data/sources/02_rooms-extended.yaml
+++ b/data/sources/02_rooms-extended.yaml
@@ -213,7 +213,7 @@
 "5505.EG.505A": { coords: { lat: 48.265385, lon: 11.668795 } }
 "5505.EG.505M": { coords: { lat: 48.265380, lon: 11.668659 } }
 "5505.EG.506": { coords: { lat: 48.265339, lon: 11.668782 } }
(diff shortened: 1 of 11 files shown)


data/sources/02_rooms-extended.yaml
```diff
@@ -314,7 +314,7 @@
 "5508.EG.808": { coords: { lat: 48.265906, lon: 11.667831 } }
 "5508.EG.899A": { coords: { lat: 48.265678, lon: 11.667768 } }
 "5508.EG.899B": { coords: { lat: 48.265773, lon: 11.667895 } }
-"5508.01.801": { coords: { lat: 48.265856, lon: 11.667780 } } # Ernst-Schmidt-Hörsaal
+"5508.02.801": { coords: { lat: 48.265856, lon: 11.667780 } } # Ernst-Schmidt-Hörsaal
```
octycs, line 317: What do you think about this change? This has so far been my main concern for updating TUMonline data, because I am quite sure it is wrong (the lecture hall really is in floor 1, and changing this to 2 would assign it to the second floor, also the map as soon as there is one).

CommanderStorm: If we keep it as is, the new TUMonline integration will fail for this room, so being consistent with TUMonline is more of a concern for us than being correct.

octycs: The best compromise I can think of at the moment would be to take over the 'wrong' IDs from TUMonline as our own IDs, but patch the room names. E.g. like this:

```suggestion
"5508.02.801":  # ID is probably wrong in TUMonline, but we need to use it for compatibility
  name: "5508.01.801 (Ernst-Schmidt-Hörsaal)"
  coords: { lat: 48.265856, lon: 11.667780 } }
```

This would should make it possible to find the room with both IDs via search, even though it would be irritating. We could further add a warning to the entry page.

octycs:
> Do you know what is the matter with this parameter?

The problem was that we need to use the `getBuildingPlacemark` instead of the `getRoomPlacemark` endpoint if we use the `b_id` parameter instead of `roomid`

octycs:
On another note, I had the issue that this import fails: https://github.com/TUM-Dev/navigatum/blob/main/data/external/main.py#L3

Usually I used to run `python main.py` in the `external` directory (we still have that in the doc), but's failing since #166 (I didn't test downloads back then). This way I need to add the parent directory to the python path (`PYTHONPATH=$PYTHONPATH:.. python main.py`). How did you run it?

data/sources/02_rooms-extended.yaml
CommanderStorm: Fun Fact:
Last time I talked to Philipp von Loewenfeld (The person who maintains the Physics roomfinder) he complained, that the ZA4 gave themselves some “unified guidelines” (temporarily breaking his roomfinder in the process)

CommanderStorm:
> How did you run it?

Pretty similar:
image

I just clicked on the arrow symbol.
image

This is the same way we import in any other file, so I must have thought that this should be respected...

I am sorry, that this broke your workflow. We can revert this change if you want.

octycs:
> We can revert this change if you want.

No I think it is okay, I just added it to the Readme

PR #189, opened 2022-08-15
Add possibility to add comments to entries

octycs:
This adds a very simple, multilingual comment field for entries. I don't expect that we will use this often, but it allows to avoid cases where users could be irritated. The background is the probably wrong ID in TUMonline for MW1801 (#187).

In the Webclient:
image

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1333,6 +1333,8 @@ components:
               $ref: "#/components/schemas/Coordinate"
             maps:
               $ref: "#/components/schemas/Maps"
+            comment:
+              $ref: "#/components/schemas/Comment"
           required:
             - id
             - type
@@ -1428,6 +1430,19 @@ components:
         - rank_combined
         - rank_type
         - rank_usage
+    Comment:
+      type: object
+      description: A comment to show to an entry
+      properties:
+        de:
+          description: German version of the comment
+          type: string
+        en:
+          description: English version of the comment
+          type: string
+      required:
+        - de
+        - en
     TokenRequest:
       type: object
       properties:
--- a/webclient/src/views/view/view-view.inc
+++ b/webclient/src/views/view/view-view.inc
@@ -181,6 +181,12 @@
         >
           ${{_.view_view.msg.no_floor_overlay}}$
         </div>
+        <div
+          class="toast"
+          v-if="view_data.comment && view_data.comment.${{_lang_}}$"
+        >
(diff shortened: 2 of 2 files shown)


openapi.yaml
```diff
@@ -1333,6 +1333,8 @@ components:
               $ref: "#/components/schemas/Coordinate"
             maps:
               $ref: "#/components/schemas/Maps"
+            comment:
```
CommanderStorm, line 1336: I dont have a strong oppinion on this, but is this not kind of an `prop`?
Shouldn't it be loated there?

openapi.yaml
```diff
@@ -1428,6 +1430,16 @@ components:
         - rank_combined
         - rank_type
         - rank_usage
+    Comment:
+      type: object
+      description: A comment to show to an entry
```
CommanderStorm, line 1435:
```suggestion
      description: A comment to show to an entry. It is used in the rare cases, where some aspect about the rooom/ translation are misleading.
```

webclient/src/views/view/view-view.inc
```diff
@@ -181,6 +181,12 @@
         >
           ${{_.view_view.msg.no_floor_overlay}}$
         </div>
+        <div
+          class="toast"
```
CommanderStorm, line 185: I would think, that for the cases we need, a `toast-warning` would be better. 
What do you think design-wise?

CommanderStorm (approved):
Overall this can be merged.
I have made some minor inline comments.

We probably need some support on the data side, or is this already there? If yes, how should I add this in https://github.com/TUM-Dev/navigatum/pull/187

webclient/src/views/view/view-view.inc
octycs: My idea was that comments can be warnings, but also just informative. That is why I would prefer a neutral toast.

octycs:
> We probably need some support on the data side, or is this already there? If yes, how should I add this in #187

No, this PR only adds the possibility to provide comments. You can easily add it e.g. in the `02_rooms-extended.yaml`:
```yaml
<id>:
  props:
    comment:
      de: "Kommentar"
      en: "Comment"
```

openapi.yaml
octycs: `props` is fine as well I think. I also don't have a strong opinion on this, maybe it's a bit more structured this way.

Issue #168, opened 2022-07-25
[Bug] API Localization


CommanderStorm:
**Describe the bug**
Currently, our API does not localize Strings.

**To Reproduce**
1. Request `https://nav.tum.sexy/api/get/mi`
2. See `props` like `Raumkennung`, `Adresse`, ...
3. See `type_common_name` like `Gebiet / Gruppe von Geb\u00e4uden`

**Expected behavior**
At at least partially localized result, to make things like https://github.com/TUM-Dev/Campus-Android/blob/a143a0cb268347bf6b8ea537083434a79754dcc2/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/roomfinder/NavigationDetailsFragment.kt#L161 something that is not needed

octycs:
This is something I also though about, but didn't have an idea that I liked so far. What I thought about:

- Should we export two differently localized versions of the data (two JSONs), or just one JSON that has both strings for each string. The first option appears more inefficient in terms of storage, but would be simpler to implement in the server (e.g. two tables in the database). A hybrid option could be to export one JSON, but the `load_api_data_to_db.py` scripts splits it into two tables.
- How should we handle localization search? Two different indices, or just one with strings in both languages. I think the search results should be independent of the language setting. So the second approach is probably better (one index, but searchable with keywords in different languages). We would need to try out how we can achieve this without too much performance decrease.
- Should we localize names? E.g. "Rudolf-Diesel-Hörsaal" into "Rudolf-Diesel lecture hall", or "Nordgelände" into "North Campus".

A first step could be the info tables (like in the PR you linked above), because this does not affect search.

CommanderStorm:
I will get on this next week, once my exams are done

PR #190, opened 2022-08-15
Add possibility to add links to entries and add some links

### LINKED ISSUE #176: [Feature] Linking to IRIS room usage data
Our data is currently quite static, but we could at least link to real time information if there are websites that provide them.

A link in the info card to https://www.devapp.it.tum.de/iris/app/ for the relevant buildings (and maybe rooms).

We could try to directly fetch the data from IRIS into NavigaTUM. But this could be a feature in the future, linking is easier to implement and could already be helpful.

**Additional context**
There are also other helpful sites to link to:
- https://www.ub.tum.de/arbeitsplaetze-auslastung
- https://tum-dev.github.io/eat-api/#!/en/mensa-garching
- https://mensa.liste.party/

octycs:
Resolves #176 in the simplest way. What is not included is the links to the library occupation, because we actually don't have entries for some of the libraries (They are usually a group of several rooms).

We still don't really have API localization (see #168), but I thought it would make sense to make the source data already multilingual. This way the data is already there.

Screenshot:
image


data/processors/export.py
```diff
@@ -87,7 +87,7 @@ def export_for_api(data, path):
         if "roomfinder_data" in export_data[_id]:
             del export_data[_id]["roomfinder_data"]
         if "props" in export_data[_id]:
-            to_delete = list(filter(lambda e: e != "computed", export_data[_id]["props"].keys()))
+            to_delete = list(filter(lambda e: e not in {"computed", "links"}, export_data[_id]["props"].keys()))
```
CommanderStorm, line 90: I always find these filter expressions really hard to parse. What do you think?
```suggestion
            to_delete = [e for e in export_data[_id]["props"].keys() if e not in {"computed", "links"}]
```

openapi.yaml
```diff
@@ -907,6 +911,33 @@ components:
...
+        url:
+          type: object
+          properties:
+            de:
+              type: string
+            en:
+              type: string
+          required:
+            - text
+            - url
```
CommanderStorm, line 937:
```suggestion
            - de
            - en
```

openapi.yaml
```diff
@@ -907,6 +911,33 @@ components:
...
+        text:
+          type: object
+          properties:
+            de:
+              type: string
+            en:
+              type: string
+          required:
+            - text
+            - url
```
CommanderStorm, line 927:
```suggestion
            - de
            - en
```

CommanderStorm (approved):
Overall, this is a good addition. I have made some minor comments, but otherwise this can be merged

From an API Perspective, I would prefer it, if we were to separate the de/en on the server and just serve what is appropriate...
I can work on this, once my exams are done... 😅

octycs:
> From an API Perspective, I would prefer it, if we were to separate the de/en on the server and just serve what is appropriate...

Yes, a general support of multiple languages in the API would be better ... if you want work on this that would be great

PR #191, opened 2022-08-15
Fixed api loading indicator

CommanderStorm:
before:
image

after:
image

Notie, that the indicator shifts around a bit on load, which I cannot debug.

### DIFF
--- a/webclient/gulpfile.js
+++ b/webclient/gulpfile.js
@@ -31,6 +31,7 @@ import fs from "fs";
 import merge from "merge-stream";
 import dartSass from "sass";
 import path from "path";
+import postcssRemoveDeclaration from "postcss-remove-declaration";
 import postcssPrependSelector from "postcss-prepend-selector";
 import source from "vinyl-source-stream";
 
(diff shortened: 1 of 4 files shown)


webclient/gulpfile.js
```diff
@@ -642,13 +643,26 @@ gulp.task("map", gulp.parallel(copyMapCSS, copyMapJS));
 
 // --- api-visualiser (currently swagger-ui) Pipeline ---
 function copyApiCSS() {
+  // swagger-ui has its own loading button
```
octycs, line 646:
```suggestion
  // swagger-ui has its own loading spinner, but it is apparently broken if we included it
```

octycs (approved):
If I understand correctly what you mean it is as far as I know actually a different effect (maybe a bug). There are two stages for the [loading indicator](https://github.com/TUM-Dev/navigatum/blob/main/webclient/src/index.html#L240) on initial pageload:
- Before vue is loaded. Until then `v-cloak` is not set and the loading indicator is visible
- When vue is loaded, but the view isn't, the `v-cloak` would normally hide the loading indicator, but it is manually added on every navigation [here](https://github.com/TUM-Dev/navigatum/blob/main/webclient/src/core.js#L378), also on pageload

Somehow there is a shift between the two stages, i.e. when vue is loaded and starts changing things on the page.

Issue #197, opened 2022-08-21
Infrastructure/ Reliability


CommanderStorm:
If you want to work on one of the issues below, please create a separate issue to discuss the details via the button next to the task.

- [x] Builds
   - [x] Revert the change from `nginx` to `unpriviliged-nginx`, as the latter causes some not debuggable reliability problems
   - [x] add `maptiler/tileserver-gl` to our deployment (cf #39)
  - [x] make `navigatum-mieli-search` build its database in docker and not on-deployment
- [ ] Infrastructure
   - [x] document used tooling
   - [x] migrate to a HA-Kubernetes cluster
     - [x] get new nodes
     - [x] get a Loadbalancer
     - [x] #317 
     - [x] document setup instructions
     - [x] revert https://github.com/TUM-Dev/NavigaTUM/commit/19220e66b988890f943e7f4cef5e5b8e1654ff79 once more resources are available
   - [x] Fix certmanager procuring certs automatically
   - [ ] Add on-prem GitHub actions for faster builds (we can throw more resources at the problem than GitHub)
- [x] Observability
   - [x] add inital monitoring via prometheus
   - [x]  add inital alerting via alertmanager
   - [x]  add inital graphing via graphana
   - [x] #440
   - [x] logging
     - [x] reduce logging noise from health checks on Rendertron
     - [x] reduce logging noise from health checks on meilisearch
     - [x] use a log-aggregator
- [ ] Workflow
  - [x] add feature-based environments (every PR gets its own deployed instance for staging changes on)
  - [x] document exact environment (supporting infrastructure) more clearly
  - [x] #111
    - [x] add a config to ImageBot, that forbids changes to svg s and revert changes to them (cf #193)
    - [x] Fix schemathesis config
  - [ ] #141

CommanderStorm:
We are currently migrating to a bigger server cluster. This will take some time.
Until then, we may not be 99.99% reliable, but [drop as low as 99%](https://stats.uptimerobot.com/xBGmxHLMYg).

The main issue is, that we are very memory constrained due to high (bot-)traffic

PR #200, opened 2022-08-25
Vue3 Migration

### LINKED ISSUE #33: migrate webclient to Vue-3
Currently the webclient works on vue-2.
Vue-3 is out, so we should look into adopting this.
The migration guide can be found here: https://v3.vuejs.org/guide/migration/introduction.html

### LINKED ISSUE #213: Firefox esr is not supported [Bug]
**Describe the bug**
My os comes with the latest version of firefox esr (91.13.0esr) which was released in august: https://www.mozilla.org/en-US/firefox/91.13.0/releasenotes/ apparently the next esr release will be a 102.x upgrade but I suppose it might take a while to land. 
With this version I get the error `ReferenceError: structuredClone is not defined` when opening navigatum. Presumably because the browser doesn't support some feature you use.

**To Reproduce**

1. Use Firefox 91.13.0esr
2. Open any room in navigatum

**Expected behavior**
The page should be displayed as intended.

**Screenshots**
image

**Desktop (please complete the following information):**

- OS: Debian GNU/Linux 11 (bullseye)
- Browser: firefox
- Version: 91.13.0esr (64-bit)

CommanderStorm:
This PR migrates the frontend to Vue3:
vue-transition-1024x521-1

# The good part:
- The new build system is HMR ready ⇒ Changes show up automatically in about 50ms, without any user input.
- The new buid system includes out of the gates multiple `.env`/request-proxying. ⇒ no more `config.js`, or changes, that get accidentally committed.
- Performance seems to be on par with the old version from what I can initially see. However, I have not yet looked into the details page. This is where I expect to see the most difference in performance (some aspects positive, some negative)
- Instead of having one large view, I have split features into smaller components if I expect that this improves maintainability.

# Caveats:
- I was unable to find a 1:1 replacement for the current **18n** approach. I will have to look further into this
- The current approach to the **dark mode** does not translate well. I have to look into this further…
- I will have to look into browser support and **polyfills**
- Somewhere in the middle my commit messages are meh, but I am not rebasing git, because of the likelyhood new errors would be introduced.

# Things, this PR won't fix:
- Until now the frontend was untested, while there is a large potential there, I think, this should be another PR.

# How to test?
```
cd server
cargo run
```
```
cd webclient
npm run dev
```

# To-dos:

- [x] initial migration
- [x] Migrate `/`  (“main”)
- [x] Migrate `/api`
   - [ ] apply postcss stlyle transforms for the assets
- [x] Migrate `/about/:id`
- [x]  Migrate `/search`
- [x]  Migrate search bar
- [x]  Migrate `/view/:id` (“details”)
   - [x] Test that every subfeature works as expected
- [x]  Migrate feedback
   - [ ] Test that every subfeature works as expected
- [x] initial deployment (vue3.nav.tum.de 😉)
- [ ] tune the performance
   - [x] chunking-strategy
   - [x] look if tree shaking is happening
   - [ ] re-check if caching works as expected
- [x] update the documentation
- [ ] re-add a sitemap
- [ ] look into polyfills

Resolves #33 
Resolves #213 (as this includes proper polyfills and does not depend on structuredClone)


CommanderStorm:
blocked by #317 and #197

CommanderStorm:
Performance seems to be okay:
image

SEO Scores are lower, as I have not integrated the meta tags into this PR

TODOs (imo these can be different PRs):
- [ ] integrae meta tags
- [ ] #474 
- [ ] Sitemaps
- [ ] Polifills (testing on older browsers and seing if the current configuration holds)

octycs:
Thank you for all the effort!

Because it's a lot of changes I'll probably do a review more by looking for bugs while using the website instead of focusing all on the code. I agree the TODOs can be a separate PR.

Just as a first thing I noticed: You have put the SCSS sources in `assets`. Wouldn't it be more common to put them in `src` directly?

webclient/src/components/ShareButton.vue
```diff
@@ -0,0 +1,59 @@
...
+      {{ $t("view_view.header.external_link.other_app") }}
+    </a>
+    <strong>{{ $t("view_view.header.external_link.share") }}</strong>
+    <button class="btn" @click="shareLink" v-if="browser_supports_share">
+      {{ $t("view_view.header.external_link.share_link") }}
+    </button>
+    <button
+      class="btn"
+      @click="copyLink()"
+      v-html="copyCurrentLink(copied) ? $t('view_view.header.external_link.copied') : $t('view_view.header.copy_link')"
```
octycs, line 56: This causes the current URL to be copied to the clipboard every time you go to a details page, because it runs on setup

CommanderStorm:
> Because it's a lot of changes I'll probably do a review more by looking for bugs while using the website instead of focusing all on the code. I agree the TODOs can be a separate PR.

Too be honest, we wont do something like this again, this shoud be split up in multiple PRs.
Live and learn 

> Just as a first thing I noticed: You have put the SCSS sources in `assets`. Wouldn't it be more common to put them in `src` directly?

I don't have a strong opinion on this. I just put it there, because it is a commonly used asset and this seemed the right place to put it...
Given that most SCSS lives in the Components, this is probably not nessesary, but I can do it if you'd prefer.

CommanderStorm:
User testing on a mac has revealed these shortcomings:
- buildings summary has an infinite loading bar
- searching does not work
- en/dark buttons don't work
- feedback type selector is bigger than on nav.tum.de

octycs:
With the `resizeObserver` I hope the scrolling problems are sufficiently fixed now. There is still a slight glitch when navigation with the browser forward button from a page at position >0 to a page scrolled to 0. In this case it scrolls to a slightly lower position first until the component is updated. I didn't investigate this, because it might be less of a problem as soon as fading is implemented again.

I tried to implement fading by the way described in [the vue-router documentation](https://router.vuejs.org/guide/advanced/transitions.html#forcing-a-transition-between-reused-views), but it causes several bugs (probably due to some async problems). We can, in order to get this finally forward, make this a new PR.

One thing I noticed for navigation is that the page is now reloaded when moving from back from an external page (Try going to [MI](https://nav.tum.de/building/mi), click on the IRIS link and then back. This is on Firefox/Chrome for me). I don't like it too much, but it's comparatively minor and probably not worth it to investigate or trivial to change.

The only thing that we might want to address before merging is this bug I can see in Chrome only (icon line widths scaled by a factor of 2):
image
Can you reproduce that?

CommanderStorm:
> The only thing that we might want to address before merging is this bug I can see in Chrome only (icon line widths scaled by a factor of 2): image Can you reproduce that?


I cannot reproduce this. This icon looks identical between Firefox/Chrome and [nav.tum.de](https://nav.tum.de)/[pr-200.nav.tum.de](https://pr-200.nav.tum.de)
image
image

octycs:
> I cannot reproduce this.

Okay, then it's probably a bug of my local Chrome version

CommanderStorm:
> Edit: I just noticed another bug: Because of the `lang="de"` in `index.html`, the language used for fetching data is always `de`

Should we ignore this until we have SSR ready, or change this in the language switcher?

Issue #39, opened 2022-02-09
[Feature] improved loading times for maps


CommanderStorm:
**Is your feature request related to a problem? Please describe.**
Currently our loading times for the interactive maps are a bit slow. We should look into this.

**Describe the solution you'd like**
Either (I am ushure if those two aproaches are compatible, this would be something to try out):

# WebWorkers
- What we might want to look into some time is using WebWorkers for (pre)caching: https://docs.mapbox.com/mapbox-gl-js/api/properties/#prewarm

# Static Images
- Also, we could try some time using the Static Images API (It might be possible to use https://github.com/jperelli/osm-static-maps for a self-hosted version): https://docs.mapbox.com/help/tutorials/improve-perceived-performance-with-static/  
This is on the Backog until I upgrade my servers storage in the summer

**Possible Tools:**
- [tilemaker](https://tilemaker.org/) for .mbtiles generation (only needed once in a while)
- [mbtileserver](https://github.com/consbio/mbtileserver) for serving

There are a lot more similar projects (just to dump them here): 
- [GeoServer](http://geoserver.org/), 
- [baremaps](https://github.com/baremaps/baremaps), 
- [tegola](https://github.com/go-spatial/tegola)).

**Describe alternatives you've considered**
Stick with the current loadingtimes, as they are, at least for the current version, enough

CommanderStorm:
I have some news considering this issue:
- [onthegomap/planetiler](https://github.com/onthegomap/planetiler) creates a mbtile file in <10h on a Laptop
- [maptiler/tileserver-gl](https://github.com/maptiler/tileserver-gl) can render this file to multiple formats utilizing a custom style

are the best combination from experimenting around a lot.

I am unsure if we should switch to static images.
The main issue is that I don't like how the static images are blurry when zooming and that there appears to be no map implementation supporting both vector and static images. 
I don't think, that adding the complexity of another map just for this feature is worth it. 

Furthermore, there seem to be some strange stuttering issues when loading new tiles.
This could be because of `tileserver-gl`'s preview, or because of some underlying issue for `tileserver-gl`.
The best course of action is, probably, to add this as a (unofficial) deployment and see if Leaflet solves this issue. 

If you (@octycs) are in the MI sometimes in the next few days, I can also show to you what I mean…
I will be there because I have to work for a course's final submission 😆

[...]

PR #201, opened 2022-08-26
Added a Tileserver

CommanderStorm:
This PR adds a Tile server to our deployment.
Currently, it only serves details for `europe`, but could theoretically serve `planet`. Serving `planet` would require a hardware upgrade, which costs [ca. €3/Month more](https://www.netcup.de/bestellen/produkt.php?produkt=2957), so I would avoid it if possible.

Mainly, this has two advantages:
- `Mapbox` has maximum requests per month, if we want to stay free. After this limit, they are costly. If we get more than 50k page loads per month, this is a fallback.
- This allows us to add more detailed preview images, which can actually show the room 😉.

Just as Rendertron, it is configured as a public facing service. (available at `/maps/`) initially.
We can reconfigure it, if necessary.

Adds a feature mentioned in #39
Solves one of the bullet points mentioned in #197


octycs (approved):
Really nice! I think `europe` should be fine :)

PR #198, opened 2022-08-22
Website/API Localisation for `/api/get/:id`

CommanderStorm:
- [x] Migrated `/api/get/:id` to include the `lang` cookie as an indicator for which langugage should be delivered
- [x] Migrated the webclient to the new API Scheme for `/api/get/:id`
- [x] Added a way to easily localize strings in the data preprocessor

partially resolves #168

PR #205, opened 2022-09-07
Better images for previews

CommanderStorm:
This PR adds nicer images for location previews in messengers
**This depends on #198 getting merged first**

Goal:
nav-card

Current status:
image

TODOs:
- [x] add deployment rule for routing this route and rethink, where it should live
- [x] There is some math I need to do. I need to figgure out, on which tiles I need to fetch and then combine into a larger picture, which I can then crop to the correct size.
- [x] Optimize Performance & Caching.
- [x] Document changes in openapi.yaml
- [x] Add changes to the frontend (#200 will take longer than this PR)


CommanderStorm:
A mistake which you may be able to learn from:
Dont profile in debug mode.

Release mode: 
image
Debug mode:
image

server/src/maps/mod.rs
```diff
@@ -0,0 +1,305 @@
...
+    )
+    .expect("Cannot open database");
+
+    let stmt =
+        match should_use_english {
+            false => conn
+                .prepare_cached("SELECT name,type,type_common_name,lat,lon FROM de WHERE key = ?"),
+            true => conn
+                .prepare_cached("SELECT name,type,type_common_name,lat,lon FROM en WHERE key = ?"),
+        };
```
octycs, line 44:
```suggestion
    let stmt = conn.prepare_cached(&format!(
        "SELECT name,type,type_common_name,lat,lon FROM {} WHERE key = ?",
        if should_use_english { "en" } else { "de" }
    ));
```
I would prefer it more like this, but it's not too important.

server/src/maps/mod.rs
```diff
@@ -0,0 +1,305 @@
...
+    // draw background white
+    for x in 0..1200 {
+        for y in 630 - 125..630 {
+            img.put_pixel(x, y, Rgba([255, 255, 255, 255]));
+        }
+    }
+    // add our logo so the bottom
+    let logo = image::open("src/maps/logo.png").unwrap();
+    //let logo_height = logo_width / 200.832 * 32.115;
+    image::imageops::overlay(img, &logo, 10, 630 - (125 / 2) - (logo.height() as i64 / 2));
```
octycs, line 247:
```suggestion
    image::imageops::overlay(
        img,
        &logo,
        15,
        630 - (125 / 2) - (logo.height() as i64 / 2) + 9,
    );
```
I have slightly adjusted the position of the logo:
image

octycs (approved):
Looks very nice! I've added only comments about small details.

Two typos: you often used 'koords' (e.g. in `x_img_koords`), I think it would be better to stay with the English 'coords'. Also I missed it the last time probably, but it should be `DetailsQueryArgs` not `DetailsQuerryArgs` (only one 'r').

CommanderStorm:
Thanks ❤️

PR #218, opened 2022-10-01
localised `type_common_name` and `usage`

CommanderStorm:
When browsing our site, I noticed that these fields are not localised:

Before
image

After
image

This PR fixes this


octycs (approved):
I did minor changes to the translation. If you agree, we can merge this :+1:

Issue #119, opened 2022-05-16
[Bug] Geolocate-controll styling


CommanderStorm:
**Describe the bug**
The geolocatecontroll does not respect our colorscheme.

**To Reproduce**
1. Tap the geolocatecontroll
2. See ugly cyan

**Expected behavior**
I would expect the locate me button to have TUM-relevant styling.

**Screenshots**
![image](https://user-images.githubusercontent.com/26258709/168697089-24b19c61-e80c-4dbf-bb03-68611b64906a.png)

**Additional context**
Not a high-priority bug

Frittenteller:
Can you assign me to the issue? I want to try

Marius1501:
Hello is the issue still relevant? If yes could you assign me to the issue? Thanks 👍🏻

octycs:
Yes, it's still open. Since there hasn't been any activity since some months, I think it's save to reassign to you.

Marius1501:
Do I have to change only the color of the button or do you expect something more?

CommanderStorm:
After clicking on the button, there is also an overlay in the same colour. Both should be included (if possible).
The main point why this is an issue is, that neither me nor @octycs had the time to look at which of the colors looks best in this context.

We have a wide array of colors already in use, but also these new colors TUM just dropped:
![Colours](https://user-images.githubusercontent.com/26258709/187756857-1edb2aab-23b5-4eba-8550-62b7b0bfa598.png)

(if you have a color, which looks good in this context and is not on our prevous roster, this is fine too 😉)

Marius1501:
Hello, I have an issue when I want to start gulp I get this error. 
![grafik](https://user-images.githubusercontent.com/109673982/190923244-5881015c-b3ef-4cd8-9607-eb59906facfb.png)

What am I doing wrong?

CommanderStorm:
I have just tested successfully it on my machine.

Could you include more of the errormessage?

FYI: Our Windows compatibility is probably not the best, since we both use some flavor of Linux. Sorry about that.
I am going to set up a Windows environment and try to replicate this.
In the meantime, you could try using WSL.

CommanderStorm:
I noticed that we are quite unclear at times what our exect dev environemnt is.
Sorry about that. We will improve this.

Okay. I got WSL working with a few hickups:
Follow https://learn.microsoft.com/de-de/windows/wsl/install to get WSL

Inside WSL:
- We want to update our caches:
  ```bash
[...]

PR #233, opened 2022-10-22
Changed user location dot color to #3070B3

Marius1501:
I changed the color of the user location dot to a TUM-color (#3070B3). The corresponding issue is #119.

### DIFF
--- a/webclient/src/views/view/view-view.scss
+++ b/webclient/src/views/view/view-view.scss
@@ -458,6 +458,15 @@
         }
     }
 
+    /* --- User location dot --- */
+    .mapboxgl-user-location-dot {
+        background-color: #3070B3;
+    }
+
+    .mapboxgl-user-location-dot::before {
+        background-color: #3070B3;
+    }
+
     #rooms-overview {
         #rooms-overview-select .menu-item {
             padding: 0;
--- a/webclient/src/views/view/view-view.scss
+++ b/webclient/src/views/view/view-view.scss
@@ -459,12 +459,10 @@
     }
 
     /* --- User location dot --- */
-    .mapboxgl-user-location-dot {
-        background-color: #3070B3;
-    }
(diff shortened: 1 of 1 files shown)


webclient/src/views/view/view-view.scss
```diff
@@ -458,6 +458,15 @@
...
     }
 
+    /* --- User location dot --- */
+    .mapboxgl-user-location-dot {
+        background-color: #3070B3;
+    }
+
+    .mapboxgl-user-location-dot::before {
+        background-color: #3070B3;
+    }
```
octycs, line 468:
```suggestion
    .mapboxgl-user-location-dot, .mapboxgl-user-location-dot::before {
        background-color: #3070B3;
    }
```

octycs (approved):
Thanks for your PR! I hope I don't scare you away with these comments at once: You can merge the two definitions into one by using a comma. I think it's better this way. It would also be also possible to write:
```scss
.mapboxgl-user-location-dot {
    background-color: #3070B3;
    &::before {
        background-color: #3070B3;
    }
}
```
(it's one of the cool features of Sass, but because it''s the same color I think it's better to use the comma version I think)

For variables that change with the theme we have the colors all defined in `webclient/src/variables.scss`. But the map doesn't change with the theme at the moment, so in my opinion it's not necessary to create a variable there.

octycs:
Nice :+1: I've just seen that there are now two blank lines above and below the `background-color: #3070B3;` :sweat_smile:

Marius1501:
Oh, should I delete them? 😄

octycs:
Yes, would be great. We usually format them like
```scss
selector {  // No space here
    prop: value;
    another-prop: value;
                                    // maybe a space here for readability
    final-prop: value;
}  // No space here
```

Marius1501:
Thank you! That`s my first merge! 😄

PR #243, opened 2022-10-26
Separated Coordinate Handling

CommanderStorm:
Given how many PRs with Coordinates, a better workflow for coordinates is in order.
This is one of the PRs, trying to improve this workflow.
By seperating coodinates and other data from each other, it makes editing them much more eazy by other (automated) tools


data/processors/merge.py
```diff
@@ -41,6 +41,30 @@ def merge_yaml(data, path):
...
+    # we assume it is a building id (which needs to be converted to string)
+    ids_to_fix = []  # Cannot change dict while iterating
+    for _id, _data in yaml_data.items():
+        if isinstance(_id, int) and len(str(_id)) == 4:
+            ids_to_fix.append(_id)
+    for _id in ids_to_fix:
+        yaml_data[str(_id)] = yaml_data[_id]
+        del yaml_data[_id]
+
+    return _recursive_merge(data, {_id: {"coords": val} for _id, val in yaml_data})
```
octycs, line 65:
```suggestion
    _recursive_merge(data, {_id: {"coords": val} for _id, val in yaml_data})
```
I haven't tested, but this return value isn't used, shouldn't we just call _recursive_merge (if `data` is modified in-place)

octycs (approved):

PR #255, opened 2022-10-28
Campus Weinstephan

### LINKED ISSUE #254: [Bug] WZW Bereich Mitte
Das Gewächshauslaborzentrum (und auch andere Gebäude) ist falsch verortet. Es liegt rund 2 km außerhalb von Freising.  
Ferner ist der Begriff WZW veraltet, denn das Wissenschaftszentrum Weihenstephan ist in der TUM School of Life Sciences aufgegangen.

CommanderStorm:
This PR renames the `WZW` to `Campus Weinstephan`.
The shortlink stays at `wzw`, as wzw.tum.de is the official link website of said campus.

Resolves #254

### DIFF
--- a/data/sources/00_areatree
+++ b/data/sources/00_areatree
@@ -239,138 +239,7 @@
     3901:Kiosk:
     3902:Seglerheim:
     3904:Bootshaus:
-  4:Campus Weihenstephan (Freising)|Weihenstephan:wzw[campus]
-    4001:Ruine der Korbinians Kapelle:
-    41:Gebiet 4100 Berg:wzw-berg
-      :Alte Akademie:wzw-berg-alte-akademie
-        4101:Verwaltung / BLQ Brau- und Lebensmittelqualität:
-        4102:Hörsaal- und Dekanatsgebäude:
-        4103:Sammlungsbau:
-        4105:Institutsgebäude I:
-        4106:Institutsgebäude II:
-        4107:Institutsgebäude III:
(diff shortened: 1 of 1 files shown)


data/sources/00_areatree
```diff
@@ -239,7 +239,7 @@
     3901:Kiosk:
     3902:Seglerheim:
     3904:Bootshaus:
-  4:Wissenschaftszentrum Weihenstephan (WZW Freising)|WZW Freising:wzw[campus]
+  4:Campus Weihenstephan:wzw[campus]
```
octycs, line 242:
```suggestion
  4:Campus Weihenstephan (Freising)|Weihenstephan:wzw[campus]
```
I think we should still include "Freising", because afaik it is commonly used

CommanderStorm:
I have moved some buildings around. Is this how you would have structured it? Should I Introduce a new TUM School of Live Sciences Parent Node?

octycs:
I'm not sure about the TUM SoLS parent node, but I tend to prefer avoiding it, because it doesn't have a meaningful location. We have some kind of parent node with the list on the main (root) page.

I only fixed the syntax (e.g. School of of Life Sciences Außenstellen was missing an ID), and reordered the two main nodes, so that they are roughly more sorted in the main tree.

CommanderStorm:
Can I merge this?

octycs:
Yes, in my opinion we can merge this

PR #257, opened 2022-10-28
Tumonline download

CommanderStorm:
This PR adds the latest data from TUMOnline
I noticed, that some fields are not consistently ordered (i.e. they shifted around from last to this run). This is why I added sorting for three of our files.


CommanderStorm:
@octycs I merged these changes, as there was a request for this from central IT

PR #259, opened 2022-10-29
Remove API lang cookie

octycs:
This removes the use of cookies for the API (because of the CORS issues, see #119 ). This only solves it for the API, there are still issues with the CDN though, because it doesn't send `access-control-allow-origin`.

Do you think it makes sense to keep `utils::should_use_english`?


server/src/utils.rs
```diff
@@ -1,11 +1,7 @@
-use actix_web::HttpRequest;
 use serde::Deserialize;
 
-pub fn should_use_english(args: DetailsQueryArgs, req: HttpRequest) -> bool {
-    // we calculate the language from the request by checking if either the query or the cookie are set to en
-    let cookie_en = req.cookie("lang").map_or(false, |c| c.value() == "en");
-    let arg_en = args.lang.map_or(false, |c| c == "en");
-    arg_en || cookie_en
+pub fn should_use_english(args: DetailsQueryArgs) -> bool {
```
CommanderStorm, line 3: We might want to consider to rename `DetailsQueryArgs` to something like `LangOption`
If we do it like this, we can make it more explicit, what is meant.

I also would think that we might want to change the implementation to a `impl DetailsQueryArgs` of said stuct, i.e. support a syntax like `args.should_use_english()`, as this is probably more ergonomic. 
What do you think?

webclient/src/core.js
```diff
@@ -10,18 +10,16 @@ let navigatum;
 const cachedFetch = (() => ({
   fetch: function (url, options) {
     return new Promise((resolve) => {
+      // Add language query param to the request
+      url += (url.indexOf("?") > 0 ? "&lang=" : "?lang=") + (localStorage.getItem("lang") || "de");
```
CommanderStorm, line 14: No strong oppinion, but this seems a bit eazier to parse. What do you think?
```suggestion
      const lang = localStorage.getItem("lang") || "de";
      url += url.includes("?") ? `&lang=${lang}` : `?lang=${lang}`;
```

CommanderStorm (approved):

octycs:
I noticed a bug actually that is also present at the moment. If you don't set a cookie by changing the language, the API now assumes German as default, but the webclient uses the Language header, which might set the page to be in English. So in this case you will see a mix of languages. We could solve this by parsing the Language Header in the Server, but we could also easily read it from `document.documentElement.lang`. Which one do you prefer? I would say the second option is simpler.

CommanderStorm:
I think this is an unrelated problem, or do you think otherwise?
I would make this its own PR/ issue, or is dealing with it here eazier?

octycs:
It would be easy to replace `localStorage.getItem("lang") || "de"` by `document.documentElement.lang` here to fix it, so I'd opt for that

webclient/src/core.js
octycs: I've updated it, but I'd still prefer the string concatenation. Would you agree on the current version?
```js
      const lang = document.documentElement.lang;
      url += (url.includes("?") ? "&lang=" : "?lang=") + lang;

```

CommanderStorm: Yes, we could even inline lang, since the reason I did extract it is no longer present

server/src/utils.rs
octycs: I would have to try this out, I'm not sure how to do this the best way in Rust. But if that's possible it would be more convenient.

CommanderStorm (approved):

Issue #34, opened 2022-02-06
Integrate extra information from the physics roomfinder


CommanderStorm:
as mentioned in https://github.com/TUM-Dev/navigatum/blob/main/deployment/README.md we still need to tap into the potential datasources of the physics roomfinder

CommanderStorm:
The NAT wants to shut down their Roomfinder by the end of March and institute a redirect to nav.tum.de

Given that 
- I currently am in my exam phase for the next few weeks and
- will be working on map overlays afterwards (Bachlor Thesis 😉),  

I am unsure if I can get this done in this time frame.

Do you (@octycs) have the resources to take over the work on this issue?

octycs:
Yes, since I have finished my exams recently, I can take it over

PR #263, opened 2022-10-31
Refactored Scaper

CommanderStorm:
This PR refactors the scaper from one single blob to more split up collection.
This is a preparation for #34, as this will increase the complexity of the scraper slightly and I think this way this is more managable


octycs (approved):
Yep, makes sense to split them now

PR #267, opened 2022-10-31
NAT Scaper

CommanderStorm:
This PR adds a scraper for the NAT roomfinder (originally the physics department roomfinder)
This is one PR in the quest for #34

### DIFF
--- a/data/external/scrapers/nat.py
+++ b/data/external/scrapers/nat.py
@@ -43,7 +43,7 @@ def scrape_rooms():
     logging.info("Scraping the rooms of the NAT")
     base_info = _get_base_room_infos()
     rooms = {}
-    for room in thread_map(_download_and_merge_room, base_info, desc="Downloaded nat rooms",):
+    for room in thread_map(_download_and_merge_room, base_info, desc="Downloaded nat rooms"):
         key = room["room_code"]  # needed, as room_code is removed in _sanitise_room
         rooms[key] = _sanitise_room(room)
 
--- a/data/external/scrapers/nat.py
+++ b/data/external/scrapers/nat.py
@@ -107,7 +107,7 @@ def _get_base_room_infos():
                 # there may be files which we did not download due to one error...
 
                 if dl or batch == 1:
-                    prog.update()
+                    prog.update(batch)
 
                 if not dl and batch != 1:
                     new_batch = batch // 2
--- a/data/external/scrapers/nat.py
+++ b/data/external/scrapers/nat.py
@@ -1,0 +1,169 @@
+# This script takes care of downloading data from the school of natural sciences (oiginally the physics department)
+import json
+import logging
+from multiprocessing.pool import ThreadPool
+
+from tqdm import tqdm
+from tqdm.contrib.concurrent import thread_map
+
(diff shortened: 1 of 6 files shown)


octycs (approved):
Looks nice that we have an API for that now! We might need to be careful what/how we merge into our data (especially coordinates for rooms can be building level), but that's certainly not part of this PR.

data/external/scrapers/nat.py
```diff
@@ -0,0 +1,168 @@
...
+    return rooms
+
+
+def _sanitise_room(room: dict):
+    """
+    Sanitise the room information
+    """
+    room.pop("room_code")
+
+    return room
```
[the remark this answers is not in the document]
CommanderStorm: Jea.. originally I had more logic here (like removing the loads of null values, spaces, empty dicts/lists, ..).
Then I noticed, that the other scrapers don't do sanitization, and removed this logic.
There are a few fields, which probably could use some sanitisation (if you have a look at the scraping results, there are a few fields, which are clearly not strings but int)

That is something I wanted to discuss but forgot to highlight:
Do we want to sanitise this data here, or in the merging step?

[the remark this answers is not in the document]
octycs: For me at least the approach is usually "first scrape, then process" (to avoid anything that could break the scraping, which usually takes more time. But this is not too bad here). On the other hand some basic sanitization could make sense here if it reduces the code complexity in the merging step I think.

CommanderStorm:
I have implemented some basic sanitization…
Could you re-review this part?

data/translations.yaml
```diff
@@ -151,3 +151,25 @@ Freifläche: Open air area
...
+nicht für den Vorlesungsbetrieb geeignet: Not suitable for lecture use
+Lüftung im Dauerbetrieb (relativ laut):
+  Ventilation in continuous operation (relatively
+  loud)
+Raum steht ab 1.9.2021 nicht mehr zur Verfügung.:
+  Room is no longer available from
+  1.9.2021.
+bis auf Weiteres keine Lüftung, daher nur mit offenen Außentüren:
+  No ventilation until
+  further notice, therefore only with open outer doors
```
octycs, line 175: Would it make sense to create a warning if there is no translation for a value of `bauarbeiten`?

octycs (approved):
As far as i can see looks good now. I'm just wondering why there are two coordinates. Do you have any idea?

data/translations.yaml
CommanderStorm: There is no provided translation in the API, if we scrape this, it should be apparent in commiting that this file has to be changed. (though I see a warning about an added translation as possibly a good idea (though a different PR))

The thing I worry about more, is that these translations will never go away, but this is a different issue too…

CommanderStorm:
> I'm just wondering why there are two coordinates. Do you have any idea?

Yea…
I have no idea. (Running theory: This is backwards compatibility due to a planned API change)
Luckily, they are always identical in value.

data/translations.yaml
octycs: Yeah, I also don't like these very specific translations here very much. But we can put this in another PR

PR #268, opened 2022-11-01
Improved docs

CommanderStorm:
This PR has the following goals:
- make all system dependencies clear.
- move some parts of our docs to other, more appropriate places.
- Communicate more clearly what is needed for which part.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Else you can follow the steps in the [data documentation](data/README.md).
 
 ### Server
 
-In case you do not want to work on the webclient only (i.e. skip the server + data), you can instead use our docker images to skip this step:
+If you want to work on the webclient only (and not server or data), you don't need to set up the server. You can instead either use the public API (see the [webclient documentation](webclient/README.md#Testing)) or use our ready-made docker images to run the server locally:
 
 ````bash
 ```bash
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -61,41 +62,49 @@ wget -P data/output https://nav.tum.de/cdn/search_data.json
 wget -P data/output https://nav.tum.de/cdn/search_synonyms.json
 ```
 
-Else you can follow the steps in the [data documentation](data/).
+Else you can follow the steps in the [data documentation](data/README.md).
 
 ### Server
 
-Follow the steps in the [server documentation](server/).
+In case you do not want to work on the webclient only (i.e. skip the server + data), you can instead use our docker images to skip this step:
```
octycs, line 69:
```suggestion
If you want to work on the webclient only (and not server or data), you don't need to set up the server. You can instead either use the public API (see the [webclient documentation](webclient/README.md#Testing)) or use our ready-made docker images to run the server locally:
```

I think we should make it clear here, that there are actually these two options for running locally without compiling the server. Else there's a difference between the webclient doc and this file.

octycs (approved):
Just a note: There is still `### 2. Starting the server` in server/README.md, not `### Starting the server`

Issue #220, opened 2022-10-06
[Bug] Share button not clickable


CommanderStorm:
**Describe the bug**
On IOS Mobile, the share button does not open the share menu (no pop up)

**To Reproduce**

1. Go to the details tab
2. Click on the share button
3. See nothing

**Expected behavior**
Our share menu

**Smartphone:**

- Device: iPhone 11 Pro Max
- OS: 16.0 (20A362)
- Browser: safari iOS
- Version: latest

**Additional context**
Related HTML: https://github.com/TUM-Dev/NavigaTUM/blob/adac394f28b91074e6f9c75c8ba0b7ac68b141e5/webclient/src/views/view/view-view.inc#L119
Related SCSS: https://github.com/TUM-Dev/NavigaTUM/blob/adac394f28b91074e6f9c75c8ba0b7ac68b141e5/webclient/src/views/view/view-view.scss#L62

CommanderStorm:
Reolved by #269 
Thanks @octycs ❤️

PR #269, opened 2022-11-01
Try to fix the share menu not opening on Safari

octycs:
This tries to fix the share menu not opening on safari (#220).

The menu opens when its button is focused, which usually happens when you click on it. This makes it very easy to open and especially close (by clicking somewhere else) the menu CSS only. But maybe Safari doesn't focus the button upon a click.

My problem is I don't have an iOS device available to test, so I can just guess. This patch _shouldn't_ pose a problem for other browsers, unless they have a different opinion on that `this` in the context of a click means. It doesn't affect the recent Firefox or Chromium versions. So maybe we can just merge and test.

### DIFF
--- a/webclient/src/views/view/view-view.inc
+++ b/webclient/src/views/view/view-view.inc
@@ -120,7 +120,8 @@
         <button
           class="btn btn-link btn-action btn-sm"
           title="${{ _.view_view.header.external_link.tooltip }}$"
-        >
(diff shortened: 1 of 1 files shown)


CommanderStorm:
Jea…
Testing in production is the easiest way…
 I will meet people tomorrow, who have iPhones, so I can test this on them ;)

Issue #248, opened 2022-10-27
[Feature] Introduce Ctrl + k to focus Search bar


0ql:
At any point where the search bar is visible on the page pressing `ctrl + k` focuses the search bar so you can start typing the search query instantly.

Something like this:
```javascript
window.addEventListener("keydown", (e) => {
    if (e.ctrlKey && e.key === "k") {
       e.preventDefault();
       document.getElementById("search").focus();
    }
})
```

Maybe pressing `ctrl + k` again could unfocus the search bar but that really isn't that important.

CommanderStorm:
I would argue, that we should not override the browsers default shortcuts, but use `s` or `/` like GitHub/...
But this is a minor thing.
I think it would be best to queue this after #200

0ql:
Yeah, makes sense. Or maybe if you just start typing it would auto focus the search. If there are no other key binds, that could be even more convenient.

0ql:
Closing due to https://github.com/TUM-Dev/NavigaTUM/pull/270

Issue #106, opened 2022-05-09
[Feature] Support Query-filters for `/api/search`


CommanderStorm:
**Is your feature request related to a problem? Please describe.**
We should support [google dorking](https://en.wikipedia.org/wiki/Google_hacking) style querry parameters/filters.

**Describe the solution you'd like**

Examples:

- [x] `"hs1"` search for exactly this string
- [ ] `+hs1` force this string being in the result
- [ ] `-hs1` filter this string not being in the result
- [ ] `@garching` filters for campus=garching
- [ ] `in:mi` filters, that the room/.. is contained in the `mi` 

**Describe alternatives you've considered**
Not implementing this. This is not a high priority issue.

**Additional context**
There already is partial support in some parts for some of these issues. This part does however require a lot more work.

octycs:
Point three (negative keyword search with `-mi`) is under consideration in MeiliSearch ([see their roadmap](https://roadmap.meilisearch.com/c/104-negative-keyword-search)), but won't be there soon. So we'd need to implement that manually.

Quoted search should be there already I think.

CommanderStorm:
See the testcases added in #104 for why this is written down ;)
It is actually not supported, at least for the testcases I added..

CommanderStorm:
Actually, any changes can probably be merged pretty “eazy-ish”
Anything search is just basically merged in this file: https://github.com/TUM-Dev/navigatum/blob/25ba18c19de000d2a32958b2d1af0792fe8a518a/webclient/src/components/AppSearchBar.vue
and some components are here: https://github.com/TUM-Dev/navigatum/blob/25ba18c19de000d2a32958b2d1af0792fe8a518a/webclient/src/modules/autocomplete.ts

The migration does mostly code-merges, moves and minor refactoring (despite how it looks)

PR #270, opened 2022-11-02
press `/` to focus the search bar

0ql:
Resolves https://github.com/TUM-Dev/NavigaTUM/issues/248

I figured `/` makes more sense as it more widely used and `ctrl + k`, as discussed, would override default keybindings of some browsers.


webclient/src/components/AppSearchBar.vue
```diff
@@ -7,6 +7,14 @@ import type { SearchResponse } from "@/codegen";
...
 
 export default {
+	mounted() {
+		window.addEventListener("keydown", (e) => {
+			if (e.key === "/") {
+				e.preventDefault();
+				document.getElementById("search")?.focus();
+			}
+		})
+	},
```
CommanderStorm, line 17: formatting
```suggestion
  mounted() {
    window.addEventListener("keydown", (e) => {
      if (e.key === "/") {
        e.preventDefault();
        document.getElementById("search")?.focus();
      }
    })
  },
```

CommanderStorm:
I am going to have to check if this interferes with typing in other fields.
This is going to take more than this week, as I currently am pretty buzy with Uni stuff. Sorry

webclient/src/components/AppSearchBar.vue
```diff
@@ -7,6 +7,14 @@ import type { SearchResponse } from "@/codegen";
 import { useFetch } from "@/utils/fetch";
 
 export default {
+  mounted() {
+    window.addEventListener("keydown", (e) => {
+      if (e.key === "/") {
+        e.preventDefault();
+        document.getElementById("search")?.focus();
+      }
+    })
```
CommanderStorm, line 16:
```suggestion
    window.addEventListener("keydown", (e) => {
      if (e.ctrlKey && e.key === "q")
        document.getElementById("search")?.focus();
    });
```

CommanderStorm: https://stackoverflow.com/questions/11941180/available-keyboard-shortcuts-for-web-applications
Maybe `Ctrl + Shift + F` and `Ctrl + Shift + S`  would be a viable option then..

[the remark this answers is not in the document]
CommanderStorm: > I think only choosing `s` or `/` are fine.

GitHub, for example, supports both. I don't see a downside to include `(e.key === "s" || e.key === "/")` 

> Also the placeholder text of the input bar should be changed to something like `Press / to Search`.

I don't think, that messing with the placeholder is the right thing to do.

A `placeholder` should (according to https://www.w3schools.com/tags/att_input_placeholder.asp):
> The `placeholder` attribute specifies a short hint that describes the expected value of an input field

We should probably have a question mark button in the search bar (like GitHub, but showing more functionaility) which if hovered over [opens a tooltip](https://picturepan2.github.io/spectre/components/tooltips.html).
The reason I like this solution more than your solution is, that this way we could also explain what query features we support in the future (see #106)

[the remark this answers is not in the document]
0ql: I suggest adding the functionality for the focusing itself in this PR and then we could open another for the way we want to inform the user about it. `s` and `/` or _only_ `/` is both totally fine with me, my reasoning behind only giving one functionality one key-bind is that it leaves more room for future expansion and otherwise you also introduce unnecessary redundancy.

webclient/src/components/AppSearchBar.vue
```diff
@@ -7,6 +7,18 @@ import type { SearchResponse } from "@/codegen";
 import { useFetch } from "@/utils/fetch";
 
 export default {
+  mounted() {
+    window.addEventListener("keydown", (e) => {
+  		if (
+        (e.key === "/") &&
```
CommanderStorm, line 13:
```suggestion
        (e.key === "s" || e.key === "/") &&
```

CommanderStorm:
Thank you for this contribution ❤️

Expect not to see this on mainline for a few weeks, as we still have to look into the deployment+testing for #200

PR #273, opened 2022-11-05
Robust previews

CommanderStorm:
Currently our `/api/preview/{id}` endpoint does not handle there not being a tileserver gracefully (read: it does not handle this 😉)

This PR adds a gracefull fallback for this case.
This was brought to light by https://github.com/TUM-Dev/NavigaTUM/commit/9c822cf2ba37f31a84c2b48314b8b1d14f8892f9
and the acompaiiing outage for this sub-api


octycs (approved):
Good to have a fallback :+1:

PR #277, opened 2022-11-07
Allow opening search suggestions in new tab

xarantolus:
This PR fixes TUM-Dev/NavigaTUM#275

I tested this in Firefox and Chromium on Windows, as well as Bromite (Chromium-based) and [Fennec](https://f-droid.org/en/packages/org.mozilla.fennec_fdroid/) on Android.

The click attribute has been changed to only trigger when no other modifier keys are pressed, that way browsers can use their native handling for Control+Click, Shift-Click, Right-Clicking etc. 

The link points to `/view/...` even for entries where other links would be more appropriate (e.g. `/campus/...`), but upon opening it gets corrected. This is in line with how the normal search result page links to entries.

Demo recording

### DIFF
--- a/webclient/src/index.html
+++ b/webclient/src/index.html
@@ -112,7 +112,8 @@
                         v-if="s.n_visible === undefined || (s.n_visible !== undefined && i < s.n_visible) || s.expanded"
                         class="menu-item">
                       <a v-bind:class="{active: e.id == search.autocomplete.highlighted}"
-                         @click="searchGoTo(e.id, true)"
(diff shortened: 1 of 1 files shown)


octycs:
That looks nice, thank you!

xarantolus:
Sorry, didn't see the message before :) 
I used https://gifcap.dev to record the GIF and [this software](https://github.com/bfritscher/carnac) to display keyboard & mouse events @CommanderStorm

Issue #143, opened 2022-06-11
[Feature] Site Maps Zoom


CommanderStorm:
**Is your feature request related to a problem? Please describe.**
The Site Maps are pretty small. Especially for older folks/ people with disabilitys this represents a significant hurdle. As the text on the image has a fixed size.
![image](https://user-images.githubusercontent.com/26258709/173206185-551f64a3-1ac6-4370-a95f-b4c07178dc58.png)

**Describe the solution you'd like**
For accassability it may be nice to offer an option for users to zoom into the image/pop it out like for headline images.

i.e. I would prefer a solution like this:
![image](https://user-images.githubusercontent.com/26258709/173206305-92c39509-fc3a-4696-b946-c50c4196758f.png)

**Describe alternatives you've considered**
In the long term, we are reducing our reliance on these maps anyway as they offer imo worse UX than the map, but sometimes having the extra context of a site-map may be nice.

**Additional context**
Not a active issue for the person, who suggesed this but something he found a nice to add.

Marius1501:
Hi, I would like to adopt this task. Could you please assign it to me? Thanks.

octycs:
Sure :) the only thing is @CommanderStorm is planning the migration to vue3, so I hope this doesn't conflict.

Marius1501:
This issue is already implemented, isn`t it? 😄

CommanderStorm:
> This issue is already implemented, isn`t it?

No, this is still open.
When clicking on the Site-Plans a modal should pop out/ it should be possible to zoom in, or what are you referring to?

Marius1501:
Oh yes sorry, I was confused. My problem is that I am not able to **test** my code. Because on `http://localhost:8000/build/index-view-main-light-de.html` I only get the main page with a network failure. But I need one page with the **roomfinder-map**. How can I solve this?

CommanderStorm:
> My problem is that I am not able to **test** my code. How can I solve this?

Usually, this means that the API is down.
Can you check this by navigating to `https://nav.tum.de/api/source_code
`, `localhost:8080/api/get/root` and `localhost:8080/api/search?q=mi`?

- success from `api/source_code` means, that the API is online,
- success from `api/get` means it can communicate with sqlite
- success from `api/search` means it can communicate with meilisearch

What does your web browsers console put out? Are there warnings?

Note that if you are trying to use our public API, this won't work until #259 is merged.

Marius1501:
[...]

PR #281, opened 2022-11-08
Modal for roomfinder

Marius1501:
Resolves: https://github.com/TUM-Dev/NavigaTUM/issues/143
grafik

Is the roomfinder-map sufficient or should the map-cross also be implemented to the modal?
I already tried to implement the cross but it is a bit difficult with the position because I think you have to make a function like the `loadRoomfinderMap`-function for the roomfinder modal.

### DIFF
--- a/webclient/src/views/view/view-view.inc
+++ b/webclient/src/views/view/view-view.inc
@@ -355,7 +355,7 @@
             <h5 class="modal-title">${{_.view_view.modalRoomfinder.header}}$</h5>
         </div>
         <div class="modal-body">
-          <div class="content" id="modalRoomfinderContent">
+          <div class="roomfinder-map-container">
             <img
               alt="Cross showing where the room is located on the hand-drawn roomfinder map image"
               src="<!-- @echo app_prefix -->assets/roomfinder_cross-v2.webp"
--- a/webclient/src/views/view/view-view.inc
+++ b/webclient/src/views/view/view-view.inc
@@ -359,7 +359,6 @@
             <img
               alt="Cross showing where the room is located on the hand-drawn roomfinder map image"
               src="<!-- @echo app_prefix -->assets/roomfinder_cross-v2.webp"
-              style="display: none"
               v-bind:style="{'transform': 'translate(' + state.map.roomfinder.modalX + 'px, ' + state.map.roomfinder.modalY + 'px)'}"
               id="modalRoomfinder-map-cross"
             />
--- a/webclient/src/views/view/view-view.inc
+++ b/webclient/src/views/view/view-view.inc
@@ -334,7 +334,7 @@
     </div>
     <!-- Modal-Roomfinder -->
     <div
-      class="modal modal-lg active"
+      class="modal modal-lg"
       id="modal-roomfinder"
       v-bind:class="{active: map.roomfinder.modalOpen}"
     >
--- a/webclient/src/views/view/view-view.inc
+++ b/webclient/src/views/view/view-view.inc
@@ -239,7 +239,10 @@
           <div id="interactive-map" class="loading"></div>
         </div>
       </div>
-        <a @click="map.roomfinder.modalOpen=true">
+        <a
+          v-on:click="loadModalRoomfinderMap()"
+          @click="map.roomfinder.modalOpen=true"
+        >
           <div
             class="roomfinder-map-container"
             v-bind:class="{'d-none': state.map.selected !== 'roomfinder'}"
@@ -333,7 +336,7 @@
     <div
       class="modal modal-lg active"
       id="modal-roomfinder"
-      v-if="map.roomfinder.modalOpen"
+      v-bind:class="{active: map.roomfinder.modalOpen}"
     >
       <a
         class="modal-overlay"
@@ -341,7 +344,7 @@
         @click="map.roomfinder.modalOpen=false"
       >
       </a>
-      <div class="modal-container modal-fullheight">
+      <div class="modal-container modal-fullheight" id="roomfinderModal-container">
         <div class="modal-header">
           <button
             class="btn btn-clear float-right"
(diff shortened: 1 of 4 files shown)


CommanderStorm:
> On the other hand, I would also be okay to merge without the cross implemented, I don't know how much difference it makes in terms of the Vue3 port @CommanderStorm ?

I am indifferent about this. Migrating this should be quite straightforward (I cannot guarantee commit ownership though, as all commits in #200 will have to be squashed before merging into main..)

Marius1501:
@octycs did you think of this change or did I something wrong? Because it is not working properly. The modal is now open when I leave out the `v-if`. And the cross is now in the left upper corner. Maybe the **loadModalRoomfinder** function is called in the wrong place. I thought it would be appropriate to call the function when you click on the roomfinder-map.

octycs:
Did you add the `loadModalRoomfinderMap` function into `view-view.js`? I can't see it in the diff

Marius1501:
Oh sorry. I tried the code in another directory and forgot to add the `loadModalRoomfinderMap` function.

webclient/src/views/view/view-view.inc
```diff
@@ -327,6 +332,48 @@
       </div>
       <div class="divider" style="margin-top: 10px"></div>
     </div>
+    <!-- Modal-Roomfinder -->
+    <div
+      class="modal modal-lg active"
```
octycs, line 337:
```suggestion
      class="modal modal-lg"
```
The reason it's always open is because the active is still given as a class (and it overwrites vue there)

webclient/src/views/view/view-view.js
```diff
@@ -519,6 +524,22 @@ navigatum.registerView("view", {
         );
       }
     },
+    loadModalRoomfinderMap: function () {
+      const map = this.view_data.maps.roomfinder.available[mapIndex];
```
octycs, line 528:
```suggestion
      const map = this.view_data.maps.roomfinder.available[this.state.map.roomfinder.selected_index];
```
This was a mistake. Now the function doesn't throw an error anymore at least, but the cross is still not at the right place. I think the reason for that is that the modal has always size 0x0 if it is not visible. And when this function is called, it is not yet visible (or the browser hasn't re-layouted yet). I'm not sure what the best solution here is. The easiest is to do a delay or polling until the modal is visible (or the size non-zero).

webclient/src/views/view/view-view.inc
```diff
@@ -239,31 +239,36 @@
...
-          id="roomfinder-map-img"
-        />
-        <div>
-          ${{_.view_view.map.img_source}}$: {{
-          view_data.maps.roomfinder.available[state.map.roomfinder.selected_index].source
-          }}
-        </div>
-      </div>
+        <a
+          v-on:click="setTimeout(loadModalRoomfinderMap, 1000)"
```
Marius1501, line 243: I tried to delay the function but the `setTimeout()` function is somehow not working. I want that the cross is at first not displayed and when the` loadModalRoomfinderMap`-function is called the display style of the cross will change to `block`.

Marius1501:
@octycs Now the positioning of the cross should be right. But I didn't get rid of the overflow of the cross.

Marius1501:
Ok, thank you for your help! Now everything should be right from my perspective. 😄

webclient/src/views/view/view-view.inc
```diff
@@ -327,6 +332,48 @@
...
     </div>
+    <!-- Modal-Roomfinder -->
+    <div
+      class="modal modal-lg"
+      id="modal-roomfinder"
+      v-bind:class="{active: map.roomfinder.modalOpen}"
+    >
+      <a
+        class="modal-overlay"
+        aria-label="Close"
```
CommanderStorm, line 343:
```suggestion
        aria-label="${{_view_view.modalRoomfinder.close}}$"
```

webclient/src/views/view/view-view.js
```diff
@@ -519,6 +524,25 @@ navigatum.registerView("view", {
...
+          -1023 + (map.x / map.width) * (rect.width - 65);
+
+      // We cannot use "height" here as it might be still zero before layouting
+      // finished, so we use the aspect ratio here.
+      this.state.map.roomfinder.modalY =
+          -1023 +
+          (map.y / map.height) * (rect.width - 65) * (map.height / map.width);
+    },
+    delayedLoadModalRoomfinderMap: function () {
+      setTimeout(this.loadModalRoomfinderMap, 1000);
```
CommanderStorm, line 544: @octycs as you have worked with the other kinds of timeout more than I have:
Do you think this is a good option, or should we experiment with `$nextTick`

webclient/src/views/view/view-view.scss
```diff
@@ -341,12 +341,27 @@
...
         padding: .05rem .3rem;
     }
 
+    #roomfinderModal-container {
+        padding-bottom: 4em;
+    }
+
+    #modalRoomfinder-map-cross {
+        position: absolute;
+        transition: transform 1s;
```
CommanderStorm, line 357:
```suggestion
        transition: transform .3s;
```
I dont think that this being slower is better. We should merge this with `#roomfinder-map-cross`

webclient/src/views/view/view-view.scss
```diff
@@ -341,12 +341,27 @@
...
+        pointer-events: none;
+    }
+
+    #modalRoomfinder-map-img {
+        width: 100%;
```
CommanderStorm, line 362: This element already has a width. Could you explain this addition?

CommanderStorm (changes requested):
Other than the timeout and other minor questions I think this is nearly ready to merge.

One thing: This chip is important for legal reasons
Please either add a chip like this one or add a bottom row like with the images 😉 
image
image

webclient/src/views/view/view-view.scss
CommanderStorm:
```suggestion
        width: 100%; // Without this part the image doesn't fill the modal over the whole width.
```

webclient/src/views/view/view-view.js
CommanderStorm: We will fix this in post if we notice this being an issue

CommanderStorm:
Sorry for the long wait 😅

PR #282, opened 2022-11-09
Updated Overlays for StudiTUM Garching

### LINKED ISSUE #276: [Entry] StudiTUM Garching, Keller und Ebene 03 fehlen
z. B. Lounge, Musikraum sind in diesen Ebenen enthalten

CommanderStorm:
This PR adds updated Maps for the StudiTUM Garching.
The core reasoning is, that the 3.OG is currently nor present.
During creation of this floor, I noticed that the other floors are not in the same style as the MW..
@octycs Is this change OK with you? (I can recreate your style for this Floor as well..)
Do you have any suggestions?

TODO:
- [x] Update the coordinates of this map

Resolves #276


octycs:
> During creation of this floor, I noticed that the other floors are not in the same style as the MW..
> @octycs Is this change OK with you? (I can recreate your style for this Floor as well..)

I did the different colors just to hint the floor colors of the building (you probably know). If this is rather irritating than helpful, or should be more explicit, we can switch to the default blue floor.

Honestly I would like to keep the doors. I assumed that the MW map was this way, because it would be way too tedious to draw all the doors (and other details). The Galileo EG map (I'm working on the 01 currently), as well as all the Roomfinder maps have also doors drawn.

There's one typo btw, on the bottom it's "NavigatTUM".

octycs (approved):
Can be merged now from my point

Issue #113, opened 2022-05-13
[Bug] Footer not at the bottom


CommanderStorm:
**Describe the bug**
Our footer does not stick to the bottom

**To Reproduce**
1. Go to any page on 'nav.tum.sexy'

**Expected behavior**
The footer should behave like a normal footer:
- stick to the bottom if content is less than the viewports height
- stick to the bottom (i.e. be out of view), if there is more content than the viewports height

**Screenshots** (I have changed the color, to make this issue more obvious)
![image](https://user-images.githubusercontent.com/26258709/168380934-a345ced4-2b2e-4cbe-9022-ce53df0c36eb.png)

**Extra context:**
- to fix this the following workflow is probably best: go to https://nav.tum.sexy and resolve the issue using the css editor of Firefox / other tools
- Insert the fix [here](https://github.com/TUM-Dev/navigatum/blob/main/webclient/src/main.scss) (using an PR)

Marius1501:
The footer is at the bottom. Or do I misunderstand the bug @CommanderStorm ? 😄

CommanderStorm:
No, this issue is still valid:
![image](https://user-images.githubusercontent.com/26258709/200902834-07148694-30ed-40f2-a258-246628cb70dd.png)

This issue is most prevalent on “weird” screen sizes, on some mobile devices and using some zoom levels.
If I mess with the zoom, I can fairly easily reproduce this.

PR #284, opened 2022-11-10
Made the footer stay at the bottom when resizing the window

Marius1501:
Resolves: https://github.com/TUM-Dev/NavigaTUM/issues/113
I think this change solves the problem.

### DIFF
--- a/webclient/src/main.scss
+++ b/webclient/src/main.scss
@@ -2,22 +2,11 @@
 
 /* === General === */
 html {
-    height: 100%;
-    box-sizing: border-box;
     scroll-behavior: smooth;
-}
-
-*,
-*:before,
-*:after {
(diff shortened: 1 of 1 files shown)


CommanderStorm (changes requested):
I tested this and now the footer sticks to the bottom.
The problem is, that the bottom is off-screen if too short content is present:
image

I am not a CSS wizard and have some problem understanding why you added some of the lines you did. (I can remove most and don't see an obvious visual change…)
Could you add GitHub comments, why these changes are necessary?

CommanderStorm:
Sorry to say, this is not fixed:
image
Also I noticed this:
image

CommanderStorm (approved):

CommanderStorm:
Thank you for this contribution ❤️

PR #293, opened 2022-11-15
Overlays for StudiTUM Innenstadt

CommanderStorm:
This PR adds Maps for the StudiTUM Innenstadt.

Do you have any suggestions?

Inspired by #276


octycs:
First of all thank you for the maps!

I've just opened the SVG and I feel we need some kind of a more reproducible workflow for these maps. This is what I see when I open the SVG in Inkscape:

image

The font rendering issues could have a different reason, but I'm missing the margins for example or the stairs are distorted (in the browser I get the same issues). And I can only see the third floor (SVGs don't natively support multiple pages, see the Inkscape [Multipage](https://wiki.inkscape.org/wiki/index.php/Multipage) documentation). It's not necessarily an issue for this PR though.

About the maps, I think it would be better for the "roomfinder" maps to have an aspect ratio that matches more the form of the building (not so wide, more square), because then they would be displayed larger.
Apart from that we can merge this I think. There is only still the issue with "NavigatTUM" instead of "NavigaTUM" like in #276. And I think the title is not in Cantarell:

image
(top: image, bottom: What I get with Cantarell Bold)

---

On the other hand the offical CI font of TUM was Arial. I haven't seen the new design anywhere except on tum.de, which uses Roboto embedded as external font (else it has defined ` -apple-system, blinkmacsystemfont, "Segoe UI", helvetica, arial, sans-serif`). We are currently using the same fonts on our website, just in a different order and without downloading fonts for performance reasons (`-apple-system,system-ui,BlinkMacSystemFont,"Segoe UI",Roboto,"Helvetica Neue",sans-serif`). Just for comparison:

image

I would stay with Cantarell or maybe Arial for now, because it better matches the rest of our website.

CommanderStorm:
> I've just opened the SVG and I feel we need some kind of a more reproducible workflow for these maps. This is what I see when I open the SVG in Inkscape:

Honestly, the main reason why I don't use Inkscape is not something personal, but That I could not make it work workflow wise and was unable to figure it out…
Using one tool would be advantageous.

I think the best plan would be:
- Once you are back from France, you sit me down and teach me your workflow.
- Long term, I would prefer to base these maps onto CAD data. I don't know how well this will work. We will have to see…

Until then dont know what is the best option/what options we have.

CommanderStorm:
@octycs how about we merge this and #282 until we have #403 is done?

octycs (approved):
I'm sorry for the long delay on this one (the other will follow). I'd consider it ready to merge now.

PR #306, opened 2022-11-19
Stateless feedback

### LINKED ISSUE #302: [Feature] Stateless feedback
Currently the feedback api has state on what tokens it has generated and what tokens it has not.

This causes troubles when introducing replication (for reliability reasons):
- request token from s1
  - s1 adds the token to the vector
- post feedback to s2 (because of LoadBalancer/...)
  - feedback is rejected

Accept state-inconsistency in token verification, but not in Generation:
The Tokens can be generated via https://github.com/Keats/jsonwebtoken or other concepts to keep the state in the Frontend instead of the server. We still need to have a counter on how many tokens have been generated, but having more tokens than we should is less of an issue imo. 

Issue with this:
This means that tokens could be used multiple times until all servers have exhausted, or the token is no longer valid.
We could have a vector for accepted tokens, so a token is at most accepted replica times across the system

I would prefer option 2, as it is easier to implement and test.

We could introduce a consistent, statefull token generator/verificatior. 
This would be a better choice if potentially having too much feedback is an issue.

**Additional context**
Give what dependencys Feedback hast, I dont know If we should make this its own microservice.

CommanderStorm:
This PR makes our Feedback system stateless.
This is a critical step in migrating to a more replicated system.

Resolves #302


server/src/feedback/tokens.rs
```diff
@@ -0,0 +1,162 @@
...
+
+use serde::{Deserialize, Serialize};
+use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
+
+use crate::feedback::{AppStateFeedback, TokenRecord};
+
+// Additionally, there is a short delay until a token can be used.
+// Clients need to wait that time if (for some reason) the user submitted
+// faster than limited here.
+const TOKEN_MIN_AGE: usize = 10;
```
octycs, line 15: Do you think we need to decrease this value (to 1-2s maybe)? The idea was originally that you very likely won't do reasonable feedback within less than 10s, but now that you can edit coordinates, the feedback form opens only after selecting a coordinate.

octycs (approved):
Stateless feedback is very good, thank you! I haven't tested, but cannot see any issues.

server/src/feedback/tokens.rs
CommanderStorm: I don't see an immediate need:
- for the user this presents as a bit of extra wait time, but is not workflow breaking
- I timed one of my flatmates on this:  
   if you actually read the text, understanding the modal takes about this time
  (I think people read this modal, since we get delete after processing requests and people do need to check the GitHub box)

The real issue is: Why do we have a `TOKEN_MIN_AGE` in the first place? From a security perspective, I don't see a immediate upside.

octycs: No, it doesn't directly help to do rate limiting (because clients can technically generate a new token before the previous one was used). The idea was that I couldn't see a case where a feedback token is requested (i.e. the form opened) and immediately used. Reasonable/helpful feedback that we want would take people >=10s to write and send. This was intended as some kind of spam protection (of bored humans or of bots pressing wildly on buttons; we haven't seen that yet though).

CommanderStorm: > This was intended as some kind of spam protection (of bored humans or of bots pressing wildly on buttons; we haven't seen that yet though).

I am not convinced that bored humans are a real threat.
Bots are not an issue. Most bots are not even smart enough to execute JS...

It does not hurt us and does not add significant complexity, imo ⇒ we should keep it and lower the limit to 5s.
What do you think?

CommanderStorm:
```suggestion
const TOKEN_MIN_AGE: usize = 5;
```

octycs: Sounds like a good idea

PR #318, opened 2022-11-28
[Cal-5] Added the calendar-modal

Marius1501:
Closes https://github.com/TUM-Dev/NavigaTUM/issues/289

grafik

I added the modal for the calendar when clicking on the calendar icon. I used an iframe to display the TUMOnline website but the URL is not found. I don't know what's the problem because it is the same URL which is used to open the website.

### DIFF
--- a/webclient/src/views/view/view-view.inc
+++ b/webclient/src/views/view/view-view.inc
@@ -204,7 +204,7 @@
           </div>
           <div class="modal-body">
             <div class="content" id="modalCalendarContent">
-              <iframe src="view_data.props.calendar_url"></iframe>
+              <iframe src="view_data.props.calendar_url" id="calendarModalIFrame"></iframe>
             </div>
           </div>
         </div>
--- a/webclient/src/views/view/view-view.inc
+++ b/webclient/src/views/view/view-view.inc
@@ -97,8 +97,6 @@
         <a
           class="btn btn-link btn-action btn-sm"
           v-if="view_data.props?.calendar_url"
-          v-bind:href="view_data.props.calendar_url"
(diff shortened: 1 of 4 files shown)


webclient/src/views/view/view-view.inc
```diff
@@ -184,6 +182,33 @@
...
+          <div class="modal-header">
+            <button
+              class="btn btn-clear float-right"
+              aria-label="${{_.view_view.header.calendar.close}}$"
+              @click="calendar.modalOpen=false"
+            ></button>
+          </div>
+          <div class="modal-body">
+            <div class="content" id="modalCalendarContent">
+              <iframe src="view_data.props.calendar_url"></iframe>
```
octycs, line 207:
```suggestion
              <iframe v-bind:src="view_data.props.calendar_url"></iframe>
```
I think you need to use `v-bind` here

Marius1501:
@octycs Thank you very much for the hint! Now is the problem that the iframe hasn't the permission to show the website. 😞 
grafik

CommanderStorm:
image
Hmm…
Yea... It looks like the easy path via an iframe is not an option here…

I have an API for the TUMonline calendar and will ask them if I can share it openly.

CommanderStorm:
blocked until #324 is done and merged

CommanderStorm:
@Marius1501 I am closing this issue, as I don't see a way we can get this approach working.
I think a different approach is more prommissing (see #324).

Any approaches in this way should be based on #200 though, as that will be merge-ready soon and the calendar is a thing I think would be easier to develop+test this way.

PR #324, opened 2022-12-09
[Cal-5] Calendar API

CommanderStorm:
This API delivers a basic Calendar API

This is the final PR in the calendar PR-Train

### DIFF
--- a/calendar/README.md
+++ b/calendar/README.md
@@ -17,14 +17,14 @@ This is separated from the server because:
 For getting started, there are some system dependencys which you will need.
 Please follow the [system dependencys docs](resources/documentation/Dependencys.md) before trying to run this part of our project.
 
-### How to Set up the Sqlite Database (needed for the `get`, `legacy_redirect` and `preview` endpoints)
+### How to Set up the Sqlite Database
 
 #### Getting the data
 
 To populate the database, you will need to get said data.
 There are multiple ways to do this, but the easiest way is to download the data from our [website](https://nav.tum.de/).
 
-(Assuming you are in the `server` directory)
+(Assuming you are in the `calendar` directory)
 
 ```bash
 mkdir -p data
--- a/calendar/README.md
+++ b/calendar/README.md
@@ -73,7 +73,7 @@ To make sure that this specification is up-to-date and without holes, we run [sc
 python -m venv venv
 source venv/bin/activate
 pip install schemathesis
-st run --workers=auto --base-url=http://localhost:8081 --checks=all ../openapi.yaml
+st run --workers=auto --base-url=http://localhost:8060 --checks=all ../openapi.yaml
 ```
 
 Some fuzzing-goals may not be available for you locally, as they require prefix-routing (f.ex.`/cdn` to the CDN) and some fuzzing-goals are automatically tested in our CI.  
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1727,6 +1727,67 @@ components:
         - parent_names
         - props
         - ranking_factors
+    CalendarResponse:
+      type: object
+      properties:
+        entries:
+          description: The entries of the requested
+          type: array
+          items:
+            $ref: '#/components/schemas/CalendarEntry'
+        last_sync:
+          description: When the last sync with TUMonline happened.
+          type: string
+          format: date-time
+          example: '2018-01-01T00:00:00'
+      required:
+        - entries
+        - last_sync
+    CalendarEntry:
+      type: object
+      properties:
+        title:
+          description: The title of the Entry
+          type: string
+          example: Quantenteleportation
+        entry_type:
+          anyOf:
+            - type: string
+              enum:
+                - exercise
+                - exam
+                - barred
+              example: exercise
+            - type: object
+              properties:
+                Lecture:
+                  description: more precise description what kind of Lecture this is
+                  type: string
+                  example: Vorlesung mit integrierter Übung
+              required:
+                - lecture
+            - type: object
+              properties:
+                Other:
+                  description: more precise description what other event this is
+                  type: string
+              required:
+                - other
+        start:
+          description: The start of the entry
+          type: string
+          format: date-time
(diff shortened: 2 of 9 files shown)


octycs:
I was just trying to run the server and let it scrape, but it appears to fail after a short while:
```
[2023-01-07T22:18:08Z INFO  navigatum_server::calendar::continous_scraping] Starting scraping calendar entrys
[2023-01-07T22:18:17Z INFO  navigatum_server::calendar::continous_scraping] Scraped 0.09% (avg 305.3ms/key, total 9.2s) result-Stats { cnt: 30, min: 0, avg: 125, max: 2906 } in time-Stats { cnt: 10, min: 401.3ms, avg: 818.2ms, max: 3.0s }
[2023-01-07T22:18:46Z WARN  navigatum_server::calendar::continous_scraping] The following ScrapeOrder cannot be fulfilled: ScrapeRoomTask { key: "0101.01.135", room_id: 7931, from: 2022-06-10, to: 2022-06-10 }
[2023-01-07T22:18:52Z INFO  navigatum_server::calendar::continous_scraping] Scraped 0.18% (avg 742.5ms/key, total 44.5s) result-Stats { cnt: 60, min: 0, avg: 208, max: 6367 } in time-Stats { cnt: 19, min: 397.3ms, avg: 731.1ms, max: 3.0s }
[2023-01-07T22:18:59Z INFO  navigatum_server::calendar::continous_scraping] Scraped 0.27% (avg 570.6ms/key, total 51.4s) result-Stats { cnt: 90, min: 0, avg: 170, max: 6367 } in time-Stats { cnt: 29, min: 380.1ms, avg: 678.9ms, max: 3.0s }
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/calendar/tumonline_calendar.rs:154:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Do you know this problem by chance?

CommanderStorm:
> Do you know this problem by chance?
> 
> The concerned room is `0101.02.133` and the event has the singleEventID `889730446` and starts on `2022-09-28T15:00:00`. It has the status `gelöscht`, so we might want to skip these kinds of entries anyway

I have fixed this issue.
`gelöscht` or other unwanted status'es are filtered out in a central place while transitioning from the `calendar_scrape` to the `calendar` table, as this seemed cleaner. We can change this design, if you think this would be better ;)

octycs:
> We can change this design, if you think this would be better ;)

No that's good I think 👍

openapi.yaml
```diff
@@ -1673,6 +1760,50 @@ components:
...
+    CalendarEntry:
+      type: object
+      properties:
+        title:
+          description: The title of the Entry
+          type: string
+          example: 'Quantenteleportation'
+        entry_type:
+          description: The title of the Entry
+          type: string
```
octycs, line 1788: Is this correct as string? I was just able to test the API, and my responses look like this for example:

```json
{
    "title": "Experimentalphysik 3",
    "start": "2022-02-07T08:30:00",
    "end": "2022-02-07T10:00:00",
    "entry_type": {"Lecture": "Vorlesung"}
}
```

calendar/README.md
```diff
@@ -0,0 +1,97 @@
...
+ln -s ../data/output data
+```
+
+#### Setting up the database
+
+To set up the database, you will need to run the `load_api_data_to_db.py` script:
+
+```bash
+python3 load_api_data_to_db.py
+```
```
octycs, line 47:
```suggestion
```

I think this is no longer required for this readme

calendar/README.md
```diff
@@ -0,0 +1,97 @@
...
+#### Testing
+
+Of course documentation is one part of the process. If the changes are substantial, you should also run an API-Fuzz-Test:
+To make sure that this specification is up-to-date and without holes, we run [schemathesis](https://github.com/schemathesis/schemathesis) using the following command on API Server:
+
+```bash
+python -m venv venv
+source venv/bin/activate
+pip install schemathesis
+st run --workers=auto --base-url=http://localhost:8081 --checks=all ../openapi.yaml
```
octycs, line 76:
```suggestion
st run --workers=auto --base-url=http://localhost:8060 --checks=all ../openapi.yaml
```
The calendar server uses `8060` (and maybe not all checks)

server/src/main.rs
```diff
@@ -52,6 +53,11 @@ async fn main() -> std::io::Result<()> {
             .app_data(web::JsonConfig::default().limit(MAX_JSON_PAYLOAD))
             .service(source_code_handler)
             .service(health_handler)
+            .service(
+                web::scope("/api/calendar")
+                    .configure(calendar::configure)
+                    .app_data(last_sync.clone()),
+            )
```
octycs, line 60: This shouldn't be part of the 'server' main anymore, or am I misunderstanding?

octycs (approved):
Thank you for all the effort! I've got only a few comments, else I think it's ready to be merged.

server/src/main.rs
CommanderStorm: Good catch!

calendar/README.md
CommanderStorm: Since the clendar needs this database, this is sadly a requirement.
Possibly building an internal endpoint into the main api to provide a list of ids with have calendars attached would be a cleander solution, but this should be another PR

openapi.yaml
CommanderStorm: Okay, there are two ways to fix this:
I have correctly doccumented the current behaviour.
The problem is, that for some `entry_type`'s which may be interesting for the user while some others do not.
What do you think about this approach? instead?
image

calendar/README.md
octycs: You're right, I forgot that it also needs to know the entries where the calendars belong to

openapi.yaml
octycs: It makes it more cleaner I think, so I think it's better this way (as two properties of type string)

PR #334, opened 2022-12-22
[Cal-1] migration from russql to Diesel

### LINKED ISSUE #327: [Feature] Migration from `russqlite` to `diesel`
The API from [`russqlite`](https://docs.rs/rusqlite/latest/rusqlite/) not great to work with and likely error prawn.
I think it is time to switch to a real ORM.

I would like to exchange all usages of [`russqlite`](https://docs.rs/rusqlite/latest/rusqlite/) with the respective usecases from [`diesel`](https://diesel.rs/)

Sticking with [`russqlite`](https://docs.rs/rusqlite/latest/rusqlite/). While being an option, manually constructing queries and extracting from them increases the likelyhood of errors.

**Additional context**
This is becoming an issue in #324, as there more SQL is needed.

CommanderStorm:
The API from [`russqlite`](https://docs.rs/rusqlite/latest/rusqlite/) not great to work with and likely error prawn.
This PR fixes this.

resolves #327 
blocks #324

### DIFF
--- a/server/src/models.rs
+++ b/server/src/models.rs
@@ -1,6 +1,6 @@
 use diesel::prelude::*;
 
-#[derive(Queryable)]
+#[derive(Queryable, Debug, Clone)]
 pub struct De {
     pub key: String,
     pub name: String,
--- a/server/src/models.rs
+++ b/server/src/models.rs
@@ -1,22 +1,10 @@
 use diesel::prelude::*;
 
 #[derive(Queryable, Debug, Clone)]
-pub struct De {
+pub struct DBRoomEntry {
     pub key: String,
     pub name: String,
     pub arch_name: Option<String>,
-    pub type_: String,
-    pub type_common_name: String,
-    pub lat: f32,
-    pub lon: f32,
-    pub data: String,
-}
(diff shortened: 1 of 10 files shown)


server/src/models.rs
```diff
@@ -0,0 +1,25 @@
...
+pub struct En {
+    pub key: String,
+    pub name: String,
+    pub arch_name: String,
+    pub type_: String,
+    pub type_common_name: String,
+    pub lat: f32,
+    pub lon: f32,
+    pub data: String,
+}
```
octycs, line 25: I'm not sure, why do we need to separate that by language here, because `En` is never used in the code as far as I can see

CommanderStorm: Yea, that is excessive. What name would you like? 
I think `RoomEntry` could be fitting…

octycs: `RoomEntry` sounds good. Maybe also `DBRoomEntry` if we want to make clear it's from the database, but not necessarily.

octycs (approved):
Looks good to me now!

PR #340, opened 2023-01-02
[Cal-3] Calendar scraper

CommanderStorm:
This PR adds a way to generate a SQLite database and keep it in sync with TUMonline.

(This PR is part of the #324 PR Train)

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -18,9 +18,11 @@ strip = true
 
 [dependencies]
 log = "0.4.17"
-diesel = { version = "2.0.2", features = ["default","sqlite"] }
+diesel = { version = "2.0.2", features = ["default","chrono","sqlite"] }
 actix-web = "4.2.1"
-awc = "3.0.1"
+actix-rt = "2.7.0"
+rustls = "0.20.7"
+awc = { version= "3.0.1", features = ["rustls"] }
 cached = "0.41.0"
 serde = { version = "1.0.148", features = ["derive"] }
 serde_json = "1.0.89"
@@ -31,3 +33,6 @@ env_logger = "0.10.0"
 image = "0.24.5"
 rusttype= "0.9.3"
 imageproc= "0.23.0"
+chrono = { version="0.4.23", features=["serde","rustc-serialize"] }
+minidom = "0.15.0"
+rand = "0.8.5"
--- a/server/load_api_data_to_db.py
+++ b/server/load_api_data_to_db.py
@@ -6,32 +6,51 @@ from typing import Any, Union
 def add_to_database(de_data, en_data):
     """add data consisting of 2x(key, data_json, data) to the sqlite database"""
     con: sqlite3.Connection = sqlite3.connect("data/api_data.db")
-    con.execute(
-        """
-    CREATE TABLE IF NOT EXISTS de (
-        key                 VARCHAR(30) UNIQUE PRIMARY KEY NOT NULL,
-        name                VARCHAR(30) NOT NULL,
-        arch_name           VARCHAR(30), -- NOT Unique, but only used for the old roomfinder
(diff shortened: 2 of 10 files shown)


server/Cargo.toml
```diff
@@ -18,9 +18,11 @@ strip = true
 
 [dependencies]
 log = "0.4.17"
-diesel = { version = "2.0.2", features = ["default","sqlite"] }
+diesel = { version = "2.0.2", features = ["default","chrono","sqlite"] }
 actix-web = "4.2.1"
-awc = "3.0.1"
+actix-rt = "2.7.0"
+rustls = "0.20.7"
+awc = { version= "3.0.1", features = ["rustls"] }
```
CommanderStorm, line 25: This is nessesary because we use https to communicate with TUMonline

server/load_api_data_to_db.py
```diff
@@ -6,32 +6,51 @@
...
+            course_code             TEXT NULLABLE,
+            course_semester_hours   INTEGER NULLABLE,
+            group_id                TEXT NULLABLE,
+            xgroup                  TEXT NULLABLE,
+            status_id               TEXT NOT NULL,
+            status                  TEXT NOT NULL,
+            comment                 TEXT NOT NULL
+        );""",
+        )
+    # purposely, this index is only on this table and not on tmp_calendar
```
CommanderStorm, line 52:
```suggestion
    # purposely, this index is only on this table and not on calendar_scrape
```

server/src/calendar/continous_scraping.rs
```diff
@@ -0,0 +1,287 @@
...
+    while !request_queue.is_empty() {
+        let mut new_request_queue = vec![];
+        for work in request_queue {
+            let events = XMLEvents::request(client, work.clone()).await;
+
+            //store the events in the database if successful, otherwise retry
+            match events {
+                Ok(events) => {
+                    success_cnt += events.len();
+                    events.store();
```
octycs, line 260: I think `store_in_db()` would be a bit better to understand

server/src/calendar/continous_scraping.rs
```diff
@@ -0,0 +1,287 @@
...
+        assert_eq!(order.num_days(), 0);
+    }
+}
+
+async fn scrape(
+    client: &Client,
+    id: (String, i32),
+    from_year: i32,
+    year_duration: i32,
+) -> Result<usize, usize> {
```
octycs, line 246: Do I understand it correctly, that `Err` is only used to indicate that a retry was necessary, but not that the scraping failed? Maybe a custom `ScrapeResult` struct would be better then

server/src/calendar/continous_scraping.rs
```diff
@@ -0,0 +1,287 @@
...
+            .collect::<Vec<(String, i32)>>(),
+        Err(e) => {
+            error!("Error requesting all ids: {:?}", e);
+            vec![]
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct ScrapeOrder {
```
octycs, line 141: I'm not sure if I understand the name of this class. Do you mean "order" like an order in a restaurant or like the order of a sequence. Maybe `ScrapeTask` is better

CommanderStorm: True. Probably the explicit `ScrapeRoomTask` is even better.

server/src/calendar/continous_scraping.rs
CommanderStorm: agree

server/src/calendar/continous_scraping.rs
CommanderStorm: fully agree

octycs (approved):

PR #355, opened 2023-01-17
[NAT-2] Extra/Explanatory information field

CommanderStorm:
Related to #34 

- Added an extra info filed to the information card
- This card looks like this: 
  image

## How to test this PR

1. Launch the webclient
2. open the vue console
3. navigate to the details view
4. under `view_data.props.computed[0]`: add `extra: {}`
5. add `view_data.props.computed[0].extra.header: "This is the Header of the info field"`
6. add `view_data.props.computed[0].extra.header: "This is the body of the info field"`

## How has this been tested?

- visual checking if this looks good

## Checklist:

- [x] I have run the linter

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -418,6 +418,10 @@ paths:
                           text: 'Boltzmannstr. 3, EG, 85748 Garching b. München'
                         - name: Sitzplätze
                           text: '522'
+                          extra:
+                            header: Genauere Angaben
+                            body: "für Prüfungen: 102 in eng, 71 in weit, 49 in corona"
+
                       calendar_url: 'https://campus.tum.de/tumonline/tvKalender.wSicht?cOrg=19691&cRes=12559&cReadonly=J'
                     ranking_factors:
                       rank_combined: 900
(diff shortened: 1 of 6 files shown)


octycs:
image

I think on mobile it's better to open a modal instead, because hover doesn't work on touch. And also sometimes the popover is clipped to the side.

CommanderStorm:
I dont like that this is another modal in the code from a code-duplication standpoint, but accept that until we get #200 off the ground this is the best we can do..

CommanderStorm:
Works as expected. From my perspective, we could merge this.

octycs:
> I dont like that this is another modal in the code from a code-duplication standpoint, but accept that until we get #200 off the ground this is the best we can do..

I also don't like the code-duplication, but from an UX perspective I think a modal is better there

octycs:
https://github.com/TUM-Dev/NavigaTUM/pull/355/commits/15b5d2c797ee13f2070298d9c032e88eed97d8d4 adds some shadow to the popover and makes its header bold:

image

https://github.com/TUM-Dev/NavigaTUM/pull/355/commits/35d34ffd99078e6ed6933f3617a11dc9de2bd5ff fixes a bug where the modal is briefly visible on pageload.

If you're okay with these changes I'd also ready to merge.

octycs (approved):

PR #362, opened 2023-01-22
[NAT-3] Nat Buildings

CommanderStorm:
Related #34 

- This adds a way to merge in nat buildings

## How to test this PR

1. Launch the system 
2. See additional buildings

## How has this been tested?

- Verified some of the new buildings are there

## TODO

- [x] collect the missing coordinates
- [x] add the collected addresses to the data model
- [x] improve the merging behaviour

## Checklist:

- [x] I have run the linter


octycs:
@CommanderStorm I can't select you as reviewer unfortunately, but this PR is ready for review. It's kind of a building list merge by hand now, because it seemed to me that the NAT building list is in some parts at least a bit outdated.

data/processors/nat.py
```diff
@@ -0,0 +1,195 @@
...
+    for _id, _data in data.items():
+        if "b_prefix" in _data and _data["b_prefix"] == building.b_code:
+            if building.b_id is not None:
+                raise RuntimeError(f"Building id '{building.b_code}' more than once in base data")
+            building.b_id = _id
+    if building.b_id is None:
+        raise RuntimeError(
+            f"Building '{building}' not found in base data. " f"It may be missing in the areatree.",
+        )
+    return building.b_id
```
octycs, line 163: Because it was very inefficient, I've simplified this in #495. Only building 1720 was concerned

data/processors/export.py
```diff
@@ -115,6 +115,8 @@ def export_for_api(data, path):
             del export_data[_id]["tumonline_data"]
         if "roomfinder_data" in export_data[_id]:
             del export_data[_id]["roomfinder_data"]
+        if "nat_data" in export_data[_id]:
+            del export_data[_id]["nat_data"]
```
CommanderStorm, line 119: given that these are now 3, I would prefer (no strong oppinion)
```suggestion
        for key in ["tumonline_data", "roomfinder_data", "nat_data"]
            if "nat_data" in export_data[_id]:
                del export_data[_id][key]
```

data/processors/nat.py
```diff
@@ -0,0 +1,195 @@
...
+        buildings = json.load(file)
+
+    # Sanity-check: Make sure that the buildings in the data are unique
+    building_ids = [b["building_code"] for b in buildings]
+    duplicate_building_ids = {b_id: cnt for b_id, cnt in Counter(building_ids).items() if cnt > 1}
+    if duplicate_building_ids:
+        raise ValueError(f"There are duplicate buildings in the data: {duplicate_building_ids}")
+
+    for building in [NATBuilding(b) for b in buildings]:
+        if building.b_code in [
```
CommanderStorm, line 44: Given that this list has grown considerable, extracting this to a configuration file is probably wise

data/processors/nat.py
```diff
@@ -0,0 +1,195 @@
...
+def _merge_building(data, building):
+    internal_id = _infer_internal_id(building, data)
+
+    b_data = data[internal_id]
+    b_data["nat_data"] = building.as_dict()
+
+    # NAT buildings are merged after TUMonline and the MyTUM Roomfinder. So if the others
+    # weren't used as sources, but the NAT Roomfinder has this building, we know it's from there.
+    base_sources = b_data.setdefault("sources", {}).setdefault("base", [])
+    if len(base_sources) == 0:
```
CommanderStorm, line 175:
```suggestion
    if not base_sources:
```

data/sources/01_areas-extended.yaml
```diff
@@ -681,6 +728,73 @@ garching:
...
+      street: "Hans-Kopfermann-Straße 1"
+      plz_place: "85748 Garching bei München"
+      source: "navigatum"
+
+origins-cluster:
+  coords: { lat: 48.26390, lon: 11.67367 }
+  osm: ["way/159713052"]
+  props:
+    address:
+      # Official address, even though not exactly where the building is
```
CommanderStorm, line 791: Should we add this as a comment in the frontend, or do you think this is probably fine as is?
I also could not get a definitive answer

data/sources/00_areatree
```diff
@@ -557,11 +578,18 @@
       6101:Heizkraftwerk:
       6102:Betriebsgebäude:
       6103:Hochspannungsgebäude:
+      6104:Feuerwehr TUM Garching (Hallen West):
       6106:Trafostation Betriebsgebäude (TBG):
       6107:Garagenbau:
-    6202:Feuerwehrwache:
+    62:Feuerwehrwache:feuerwehrwache[joined_building]
```
CommanderStorm, line 584: I am unshure if `feuerwehrwache` is a too general id

data/sources/00_areatree
```diff
@@ -542,13 +562,14 @@
       5613:Finger 13 (BT13):
     -5622:Sportanlage Garching / Dusch Container:
     5701:Munich School of BioEngineering, IMETUM:
+    5801:Leibnitz-Rechenzenturm (LRZ):
```
CommanderStorm, line 565: should we add the id `lrz`?

data/sources/01_areas-extended.yaml
```diff
@@ -733,9 +847,17 @@ frm-2:
   coords: { lat: 48.26826, lon: 11.67866 }
   osm: ["way/159713511"]
 
+5201:
+  coords: { lat: 48.26582, lon: 11.67346 }
+  osm: ["way/774866203"]  # OSM says: 5258
```
CommanderStorm, line 852: If we are confident this is the right object, we could provide a patch to OSM, how sure are you, that this match is correct?

data/sources/00_areatree
octycs: I am not sure if 5801 is the whole LRZ (my assumption is it's not). The ID `lrz` would make sense if we know there is a parent with multiple buildings belonging to it. But with just one child building I think it doesn't make sense at this moment

data/sources/01_areas-extended.yaml
octycs: It comes from the fact that the building name is `UYL` in OSM, which is also the building name from the NAT roomfinder and correct according to [this source](https://www.frm2.tum.de/fileadmin/w00bnv/www/content_uploads/experimental-facilities_online.pdf) (page 118 or 116, "Access to the MLZ"). So it depends on how sure we are about that the ID in the NAT Roomfinder is correct (and not outdated)

data/sources/00_areatree
octycs: Yes maybe `feuerwehrwache-garching` is better? Both 61.. and 62.. belong to the the Werksfeuerwehr. I'm not sure how exactly they are separated.

data/sources/01_areas-extended.yaml
octycs: The coordinates in the NAT Roomfinder do at the moment point to the building that we have selected and according to OSM it's also the one. I would leave it like this I think because I am not sure enough about how it's intended to be to write something in the Frontend. We could merge now but ask someone from there at the same time and fix if needed.

data/processors/nat.py
octycs: I've moved it to a YAML file

CommanderStorm:
(we can merge this now or with #495, whatever you like more)

Issue #352, opened 2023-01-11
[Feature] [Entry] Fachschaft BUV/BGU


Aevyz:
https://nav.tum.de/room/0101.02.157 sollte die BUV sein

https://www.fs.bgu.tum.de/de/kontakt#Raum

Das die TUM denkt, dass das ein CIP Pool ist, ist eine andere Sache...

CommanderStorm:
I have renamed you as you call yourself on your website and changed the type to "Office" (which is probably more correct)
If you think this is incorrect, dont hesitate to reply to this issue

CommanderStorm:
(just a heads-up: open the website in an inkognito tab, as this field is cached for some time)

Aevyz:
> I have renamed you as you call yourself on your website and changed the type to "Office" (which is probably more correct) If you think this is incorrect, dont hesitate to reply to this issue

Could it be, that the Search is still broken? Looking for [0101.02.157](https://nav.tum.de/search?q=0101.02.157) still results in the CIP being listed

CommanderStorm:
Good catch,
What I did not think about in #345 is that the search API also manages state.
Decoupling API-Data also means, that Search-Data has to be handled this way.

I have poked it and the symptoms (!= root cause) should be gone.

PR #367, opened 2023-01-27
Reduced sourcecode handler

CommanderStorm:

- This PR removes the source code handlers
- Reasoning:
  - For debugging, I don't consider this useful
  - Why keep something that is not useful?

## How to test this PR

1. Discuss if sourcecode is not needed instead

## How has this been tested?

- internal monolouge, that this feature is not needed

## Checklist:

- [x] I have run the linter

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -10,10 +10,14 @@ mod utils;
 const MAX_JSON_PAYLOAD: usize = 1024 * 1024; // 1 MB
 
 #[get("/api/status")]
-async fn health_handler() -> HttpResponse {
+async fn health_status_handler() -> HttpResponse {
+    let github_link = match std::env::var("GIT_COMMIT_SHA") {
+        Ok(hash) => format!("https://github.com/TUM-Dev/navigatum/tree/{hash}"),
+        Err(_) => "unknown commit hash, probably running in development".to_string(),
+    };
     HttpResponse::Ok()
         .content_type("text/plain")
-        .body("healthy")
+        .body(format!("healthy\nsource_code: {github_link}"))
 }
 
 #[actix_web::main]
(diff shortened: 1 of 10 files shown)


octycs:
Since it's GPL (as said in the doc), we don't need to have it. For me it's kind of equal whether to keep it or to remove it. It follows the idea of GPL to provide access to the source code (so it wasn't intended for debugging anyway), but the repo is also directly accessible from the NavigaTUM website. The only thing is that people without access to the deployment might not know which commit is currently active. But if you say it's rather unnecessary or even standing in the way, I am okay to remove it.

CommanderStorm:
> The only thing is that people without access to the deployment might not know which commit is currently active.

Normally I would respond that this never happens and if it does it would be my job to spot this via the alerting we have.
But #352 has shown me some gaps in my alerting for which I really dont have a good way to adress them, so you are probably right.
Thinking more about this, I think this should be combining the `health` and the `source_code` endpoint into a `status` endpoint.

server/src/main.rs
```diff
@@ -9,20 +9,7 @@ mod utils;
...
-        Err(_) => gh_base,
-    };
-    HttpResponse::Ok()
-        .content_type("text/plain")
-        .body(github_link)
-}
-
-#[get("/api/health")]
+#[get("/api/status")]
 async fn health_handler() -> HttpResponse {
```
octycs, line 13: Shouldn't this also be adapted?

octycs (changes requested):

octycs:
> Thinking more about this, I think this should be combining the health and the source_code endpoint into a status endpoint.
That's a good compromise :+1:

server/src/main.rs
CommanderStorm: Ups

octycs (approved):

Issue #130, opened 2022-05-25
Announcement issue for API changes


CommanderStorm:
This Issue tracks, **who uses the API** and **what features** you are using and what **change-timelines** ("I need 3 Weeks at minimum, because ...") you would prefer.

If you have implemented a part of the API, please note down in the comments which ones. Doing so will indicate, that we will have to keep this API in a stable state and cannot do breaking changes without the **change-timeline** you need.

We will use this Issue to note breaking API changes.

There are some things about compatibility and assumptions we are making:

- Us adding an aditional field should never bother your Application 
- Undocumented behaviour is not relied upon
- Bugs will be fixed without notice

CommanderStorm:
We have renamed `nb_hits` to `estimatedTotalHits` in https://github.com/TUM-Dev/navigatum/pull/162/commits/9564038ae53dc7e2f6b849dc3c33e3c0eb9e7bb1

Since nobody has commented on this issue, this is change is going online immediately and without any backwards compatability.

CommanderStorm:
We have added the localization of the details' endpoint via the `?lang=en` query parameter

COM8:
Implemented `api/search`, `api/get/{id}` and `/api/preview/{id}` in the [UWP TUM Campus-App](https://github.com/TUM-Dev/Campus-UWP).

CommanderStorm:
We have merged the `/health` and the `/source_code` endpoints into the `/status` endpoints
See [this PR](https://github.com/TUM-Dev/NavigaTUM/pull/367) for the full discussion.
Given that nobody has indicated that he/she is using this API, this is going into production this evening.

CommanderStorm:
After looking into what https://github.com/TUM-Dev/NavigaTUM/pull/472 needs I have come to the conclusion that this feature would be simpler if we did not have to have two details-responses (one for `/api/get/root` and one for `/api/get/{id}`).
=> `/api/get/root` is slated for removal in https://github.com/TUM-Dev/NavigaTUM/pull/1154

Given that nobody has indicated that he/she is using this API, this is going into production at the end of the week.

CommanderStorm:
`poi.nearby_public_transport.mvg` was moved into the dedicated `/locations/{id}/nearby` endpoint.
Reason:
- nobody currently uses that data and it is somewhat large => not worth transmitting for everybody

PR #369, opened 2023-01-28
Dropped UTM support

CommanderStorm:

- This PR drops the support for UTM being delivered to the client
- Main reason is, that I doubt anybody would ever need this feature (most tools support lat-lon)

## How to test this PR

1. Check if utm is really removed from API responses

## How has this been tested?

- Genereated a few responses, for which utm did not show up

## Checklist:

- [x] I have run the linter

Note, that this will need an entry in #130


octycs (approved):
Seems all fine. Yes, UTM is probably not used, and anyone can convert themselves. I also don't know why the original roomfinder used it.

PR #371, opened 2023-01-28
made sure that unicode splits dont cause unwraps

### LINKED ISSUE #370: [Bug] UTF-8 Grapheme cluster handling in `search`
Wenn ich nach "1.11 mri" suche erhalte ich einen Fehler 502.

CommanderStorm:
Fixes #370 

- This PR changes the handling of truncations which are in  the midde of clusters to be utf-8 friendly

## How to test this PR

1. `curl http://localhost:8080/api/search?q=1.11%20mri&lang=en`
2. code review

## How has this been tested?

- `curl http://localhost:8080/api/search?q=1.11%20mri&lang=en`

## Checklist:

- [x] I have run the linter


octycs:
I've added tests and changed the `unicode_split_at` function. The reason was first a bug (left is produced by the old version, it just starts two times from the first character):

```
thread 'core::search::search_executor::postprocess::tokenizer_tests::unicode_split' panicked at 'assertion failed: `(left == right)`
  left: `("Gria", "Griaß e")`,
 right: `("Grieß", " eich")`', src/core/search/search_executor/postprocess.rs:257:9
```

I think it is just a bit easier this way. The second reason is that I don't know how reliable the `width` is. Zero width characters are probably not of use, but are for example neglected by the truncate function. But I've kept the truncate function for the place where we really need to truncate.

octycs (approved):

server/src/core/search/search_executor/postprocess.rs
```diff
@@ -200,39 +202,38 @@ fn parse_room_formats(
...
-                .unwrap_or_default(),
+            parsed_aid.1,
         ))
     } else {
         None
     }
 }
 
+fn unicode_split_at(search: &str, width: usize) -> (String, String) {
+    (
```
CommanderStorm, line 231: Hmm… 
The thing I am a bit unhappy with style wise is that we produce 2x`String`, while we only need `&str`
Performance wise this likely does not make a difference, but I am going to try to make this a bit cleaner.

server/src/core/search/search_executor/postprocess.rs
```diff
@@ -244,3 +238,25 @@ fn generate_subtext(hit: &meilisearch::MSHit) -> String {
         None => building,
     }
 }
+
+fn unicode_split_at(search: &str, width: usize) -> (&str, &str) {
```
CommanderStorm, line 242: I found a way, which is imo a bit cleaner. What do you think about this approach?

octycs: Yeah that's good

octycs (approved):

PR #360, opened 2023-01-19
Fixed the roomfinder-cross-bug

Marius1501:
Fixes #323 

## Proposed Changes (include Screenshots if possible)

- Now it updates the cross position when the browser-window is resized
- For the main-roomfinder and the modal-roomfinder

At fullsize:
grafik

At smaller size:
grafik

## How to test this PR

1. Go to any roomfinder-map or to any modal-roomfinder
2. Resize your browser-window to any size
3. See that the cross stays at the same place

## How has this been tested?

- Windows 11
- Firefox Version 108.0.2 (64-Bit)

## Checklist:

- [x] I have updated the documentation / No need to update the documentation
- [x] I have run the linter

PR #372, opened 2023-01-28
Switch to MapLibre GL

### LINKED ISSUE #366: [Bug] Mapbox running out of quota
*deleted as requested by the sender*

This PR was originally "delete-after-processing". For documentation purposes the issue is kept.

octycs:
Fixes #366 

- Drop-In replacement of MapBox GL with MapLibre GL
- Only visual change is that there is no longer a MapBox logo
- MapBox removed from the privacy

## How to test this PR

1. Open the website and a room on the website, the map shows by default

## How has this been tested?

- Map display seems to work without issues
- Floor overlays work
- Selecting a coordinate on feedback works
- 3D works (noticed this the first time that you can use Ctrl to turn in 3D, see screenshot)
- Geolocation works
- Fullscreen works (one bug remaining¹)
- Resize of the window works now (one bug fixed in this PR as well²)

¹ Opening the fullscreen and then closing it again doesn't update the fullscreen icon. The issue also exists for the old mapbox version

² Fixed a bug from #360 where the map would switch to the roomfinder on window resize

image

## Checklist:

- [x] I have run the linter

### DIFF
--- a/webclient/gulpfile.js
+++ b/webclient/gulpfile.js
@@ -632,19 +632,19 @@ function copyWellKnownRoot() {
 }
 gulp.task("well_known", gulp.parallel(copyWellKnown, copyWellKnownRoot));
 
-// --- map (currently mapbox) Pipeline ---
+// --- map (currently maplibre) Pipeline ---
 function copyMapCSS() {
   return gulp
-    .src(["node_modules/mapbox-gl/dist/mapbox-gl.css"])
-    .pipe(concat("mapbox.css"))
+    .src(["node_modules/maplibre-gl/dist/maplibre-gl.css"])
+    .pipe(concat("maplibre.css"))
     .pipe(gulpif(config.target === "release", csso()))
     .pipe(gulpif(config.target === "release", rename({ suffix: ".min" })))
     .pipe(gulp.dest("build/css"));
 }
 function copyMapJS() {
   return gulp
-    .src(["node_modules/mapbox-gl/dist/mapbox-gl.js"])
-    .pipe(concat("mapbox.js"))
+    .src(["node_modules/maplibre-gl/dist/maplibre-gl.js"])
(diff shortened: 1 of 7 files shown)


CommanderStorm:
- Moment considering #373

octycs:
Best timing :joy:

CommanderStorm (changes requested):
Interesting how similar these PRs are. Yours does have a bit more engineering + bug fixing put into it ⇒ Works mostly as expected.

Bugs I encountered:
- Full-screen on mobile:
  
image

-  We seem to be requesting `Roboto` from `cloudflare`.
image
   I have tried to investigate this, and I don't know how to fix it.
   Things that we could do to address this:
   - document this behaviour in our privacy policy (not preferred)
   - I assume that adding `Roboto` as a dependency should make this work. 
   - Change the font the `tileserver` is configured to use
   What do you think is the best approach? (is this even a problem?)

octycs:
The maptiler url is also given in the `style.json`:

image

Can we just remove this definition maybe?

octycs:
> * Full-screen on mobile:

Hmm I can't reproduce it on clicking fullscreen neither in Chrome nor Firefox.
However I can reproduce it by switching to the roomfinder, resizing the window and switching back to the interactive map.

CommanderStorm:
> Edit 2: If we want to keep Roboto we can of course just mirror it ourself.

Working on providing the font ourselves. 😉

octycs:
https://github.com/TUM-Dev/NavigaTUM/pull/372/commits/a1de9cba7e6d8004e1232e6985930e616b556a65 fixes the problem with the black region for me. It feels a bit not ideal as a solution but it works at least.

CommanderStorm (approved):

CommanderStorm:
Given that this is independent of this change, I am merging it

PR #385, opened 2023-02-03
`calendar`-`data` dependency reduction

CommanderStorm:
Previously, when deploying data, a redeployment of the calendar was necessary.
This is part of the process of making this a StatefullSet

- Added a new API to the main-api (`/internal/list/ids_with_calendar`)
- Migrated the 

## How to test this PR

1. Launch server
2. in another terminal set 
   `API_SVC_SERVICE_HOST=localhost`
   `API_SVC_SERVICE_PORT_API=8080`
   run the calendar
3. See that the calendar comes up correctly.

## How has this been tested?

- as stated above

## Checklist:

- [x] I have run the linter


octycs (approved):
Disclaimer: I haven't tested myself, but I can't see any issues from the code at the moment. Is there already any application known to us depending on the calendar API already? If not we can merge and see if there's any issues I think.

Issue #199, opened 2022-08-22
[Feature] More human readable IDs in URLs


octycs:
**Is your feature request related to a problem? Please describe.**
For consistency, all buildings use the same ID as TUMonline. However in the URL it would be nicer to have human readable IDs if possible (e.g. `studitum-innenstadt` instead of `0201`).

**Describe the solution you'd like**
We could propose additional "IDs" that can be used in the URL (and only there to avoid confusion).

**Describe alternatives you've considered**
This is not a very important issue, so one alternative could be to leave this feature out. We could also just use human readable IDs directly instead of the TUM IDs, but it's not consistent and if a building changes its function (which might happen) we need to change the ID everywhere, instead of just the visible ID.

**Additional context**
For some buildings/areas such ids are already declared and parsed in the areatree (see the documentation to `visible-id` at the top of the areatree file and the discussing of the areatree in #196 )

Issue #30, opened 2022-02-04
Improve the handling of synonymous names


octycs:
Currently, some similar types of rooms have a lot of different names in TUMOnline, e.g.:

"Rechnerraum-Studenten", "Rechnerarbeitsraum", "Rechnerraum", "Computerarbeitsraum", "Computerraum", "Computerpool", "Computer Studenten", "CIP-Raum", "CIP-Pool", "Computerpool", "CAD-Pool", "EDV-Pool", "Rechner-Raum"

Currently, we have synonyms configured for a lot of them in Meilisearch, but too many synonyms slows search down. To optimize this we could reduce the number of terms already in the data processing, e.g. changing rarely used terms such as "Computerpool" to more common such as "CIP-Pool". Then we only need a synonym to map "Computerpool" => "CIP-Pool", but not the other way round.

Additionally, the server might use these synonyms to optimize the display of results, e.g. show it in the subtext.

CommanderStorm:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** enhancement
**Summary:** Reduce Meilisearch synonym-table size by normalising room-type names at data-processing time and switching the remaining synonyms to one-way mappings — **only if** the current bidirectional synonym table is actually slowing search down. Measure first.

**Current behavior:**
All room-type synonym handling lives in `server/src/setup/search_synonyms.yaml` as **bidirectional** synonym groups. The original reporter's example (Feb 2022) still applies verbatim: 13 different German names for "computer room" (`Rechnerraum`, `Computerraum`, `CIP-Pool`, `EDV-Pool`, `CAD-Pool`, …) are each mapped to all 12 others, producing a dense N×N synonym graph. Since 2022 the synonym file has grown, not shrunk — additional bidirectional groups have been added for libraries, lecture halls, meeting rooms, copy/print rooms, kitchens, cooling, waste, etc. The source room-type strings in `data/external/results/rooms_*.csv` are not normalised; the data pipeline carries the raw variants through unchanged.

**Desired behavior:**
Two paths, gated by measurement:

[...]

PR #396, opened 2023-02-06
Room names

CommanderStorm:
Makes sure that things like #394, #393, #392 don't happen

- refactored that our room patches no longer expect that the id is in the name
- removed unused methods

## How to test this PR

1. Code Review
2. Run the data pipeline

## How has this been tested?

- internal Code Review
- Ran the data pipeline

## Checklist:

- [x] I have run the linter

### DIFF
--- a/data/sources/02_rooms-extended.yaml
+++ b/data/sources/02_rooms-extended.yaml
@@ -112,8 +112,8 @@
   name: Ernst-Schmidt-Hörsaal
   props:
     comment:
-      de: Es wird auch 5508.02.801 statt 5508.01.801 verwendet, obwohl der Hörsaal im 1. Stock ist.
-      en: Also 5508.02.801 instead of 5508.01.801 is being used, even though the lecture hall is in the 1st floor.
+      de: Dieser Hörsaal ist in TUMonline 5508.02.801 statt 5508.01.801, obwohl der Hörsaal im 1. Stock ist. Aus Konsistenzgründen übernehmen wir diese ID.
+      en: This lecture hall is 5508.02.801 in TUMonline instead of 5508.01.801, although the lecture hall is on the 1st floor. For consistency, we also use this ID.
 "5510.EG.026C":
   name: IKOM-Besprechungsraum
 
--- a/data/sources/02_rooms-extended.yaml
+++ b/data/sources/02_rooms-extended.yaml
@@ -1,125 +1,125 @@
 # ========================== Innenstadt =============================
 # === Arcisstraße ===
 "2907.01.130":
-  name: "2907.01.130 (Fachschaft Lehrertum/Lehramt)"
+  name: Fachschaft Lehrertum/Lehramt
 "0501.03.173":
-  name: "0501.03.173 (Vorhoelzer Forum Foyer)"
+  name: Vorhoelzer Forum Foyer
 "0501.05.131":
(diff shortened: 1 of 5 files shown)


data/sources/02_rooms-extended.yaml
```diff
@@ -1,125 +1,125 @@
...
+  name: Rechnerhalle
 "5508.02.801":  # ID is probably wrong in TUMonline, but we need to use it for compatibility
-  name: "5508.01.801 (Ernst-Schmidt-Hörsaal)"
+  name: Ernst-Schmidt-Hörsaal
   props:
     comment:
-      de: "Es wird auch 5508.02.801 statt 5508.01.801 verwendet, obwohl der Hörsaal im 1. Stock ist."
-      en: "Also 5508.02.801 instead of 5508.01.801 is being used, even though the lecture hall is in the 1st floor."
+      de: Es wird auch 5508.02.801 statt 5508.01.801 verwendet, obwohl der Hörsaal im 1. Stock ist.
+      en: Also 5508.02.801 instead of 5508.01.801 is being used, even though the lecture hall is in the 1st floor.
```
octycs, line 116: There is unfortunately the problem, that this way we cannot patch the ID in the title anymore. If you would now search for "5508.01.801" (the old and I think more correct ID), you wouldn't find the room anymore. Not sure of a clean solution, but I think there needs to be a way to make a patch to the full / final room name

CommanderStorm: I think this should be handled via #30 or #199 and not via this mechanism
Furthermore, I am not convinced that people primarily use the IDs for finding rooms, but instead defer to using the Human name (in this case `MW1801`) so I would argue, that breaking this is nothing that should be a showstopper

The ways we could patch this involves a regex or dirty measures, which likely are going to cause problems in the future, which I would like to avoid such crude hacks. (If you think we should, we can put them in, but this should be an intentional choice…)

octycs: The best way to fix would be to fix it upstream of course. The TCA app uses the ID for search in some functions, but then it's the ID from TUMonline. I just tested it with a different floor, and my assumption is probably wrong, it could be found (at least in the way search works currently):

image

It's still that I don't like having data that is kind of wrong. At least these errors should be documented (It is at the moment, but only human-readable in the comment). Because if we don't, we'll forget and if we just take over everything as is from TUMonline, we end up with the same data quality. For example as soon as we have floorplans for the MW upper floors, it might display the wrong floor map, because it's the wrong floor in TUMonline.

So I'd say we can leave out a feature to patch the final name, but for the future (like when it comes to the floor etc.) we will need to have patching systems.

CommanderStorm, line 116:
```suggestion
      de: Dieser Hörsaal ist in TUMonline 5508.02.801 statt 5508.01.801, obwohl der Hörsaal im 1. Stock ist.
      en: This lecture hall is 5508.02.801 in TUMonline instead of 5508.01.801, although the lecture hall is on the 1st floor.
```
Do you think a comment like this could clean this up?

octycs: I think we can shorten the last sentence to "Aus Konsistenzgründen übernehmen wir diese ID." / "For consistency, we also use this ID."

PR #412, opened 2023-02-20
added an export for props.operator to the details-api

### LINKED ISSUE #410: [Form] Expose Organization Information via /api/get
`/api/get/5606.EG.037` should expose `"organization": {"name": "TUVBMPI Fachschaftsverbund Mathematik/Physik/Informatik/Chemie", "orgNr": "45500"}` or similar somehow

CommanderStorm:
Fixes #410

- Added an export for the operator

## How to test this PR

1. Recompile the data
2. See that the output of `/api/get/5508.02.801`
3. See that the API matches.

## How has this been tested?

- See if the output of the `extract_tumonline_props()` function matches the expected format+content 

## Checklist:

- [ ] I have run the linter

### DIFF
--- a/data/processors/sections.py
+++ b/data/processors/sections.py
@@ -2,7 +2,7 @@ from utils import TranslatableStr as _
 
 
 def extract_tumonline_props(data):
-    """Extracts the calendar from the tumonline data sets it to the proper value."""
+    """Extracts some of the TUMonline data and provides it as `prop`."""
     for entry in data.values():
         if entry.get("tumonline_data", {}).get("calendar", None):
             url = f"https://campus.tum.de/tumonline/{entry['tumonline_data']['calendar']}"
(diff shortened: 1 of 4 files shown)


data/processors/sections.py
```diff
@@ -1,12 +1,19 @@
 from utils import TranslatableStr as _
 
 
-def extract_calendar_urls(data):
+def extract_tumonline_props(data):
     """Extracts the calendar from the tumonline data sets it to the proper value."""
```
octycs, line 5:
```suggestion
    """Extracts some of the TUMonline data and provides it as `prop`."""
```

octycs (approved):
We actually also have a [scraper function for the orgs](https://github.com/TUM-Dev/NavigaTUM/blob/main/data/external/scrapers/tumonline.py#L195), it's just not integrated yet. From there we could also get the full name. But this doesn't need to be in this PR. We can directly merge this imo.

octycs:
Wait, I'll make a quick PR for that

Issue #403, opened 2023-02-11
[Feature] automated overlay generation


CommanderStorm:
**Is your feature request related to a problem? Please describe.**
Our overlay quality is very dependent on the building and on the data the Roomfinder has on this.

**Describe the solution you'd like**
Our overlays are currently all scraped from the Roomfinder, it should be possible to generate them automatically from CAD Data (ZA4 Provided)

**Describe alternatives you've considered**
no viable alternative

**Additional context**
This will be processed via a Bachelor Thesis

CommanderStorm:
> *This was generated by AI during triage.*

Triaged as `feature` + `ready-for-human`. Work is in progress.

**Note on framing:** the original "automated overlay generation" approach has been superseded. The current direction is to import the CAD-derived TUM building data directly into **OpenStreetMap** and render from OSM, rather than generating NavigaTUM-specific overlays. NavigaTUM then picks the data up via its existing OSM consumption path. This work is being driven by a Bachelor Thesis.

This is `ready-for-human` rather than `ready-for-agent` because the remaining work spans OSM data modelling, CAD parsing, and coordination with the thesis student — none of which an autonomous agent should run on its own.

PR #416, opened 2023-02-23
Add Roomfinder-like maps for CH

octycs:

- Add Roomfinder map  images for our current chemie overlays (mainly so that they can be directly seen in the TCA)
- This is preliminary until we have proper data (see #403)

image

## How to test this PR

1. Recompile the latest data
2. Verify that the map images are available in your local webserver
3. Open rooms where we have coordinates (e.g. 5402.01.221K)

## How has this been tested?

- Multiple rooms in different floors

## Checklist:

- [x] I have run the linter


CommanderStorm (approved):
nice

PR #423, opened 2023-03-02
Statefull calendar

CommanderStorm:
Fixes that our calendar has excessive downtime after each deployment, becase it throws away any state

the previous mechanism was to scrape into a "dev"-database and then promote this into master.

This mechanism was changed in this way:
- addidtion of the `last_scrape` field in the database (for filtering out deleted entrys)
- removal of the shared state
- switch to upsert operations instead of bulk-insert

## How to test this PR

1. Code review
2. Test in production

## How has this been tested?

- build+run, look at the database

## Checklist:

- [x] I have run the linter


server/calendar/src/main.rs
```diff
@@ -20,14 +20,17 @@ async fn health_status_handler() -> HttpResponse {
...
 async fn main() -> std::io::Result<()> {
     env_logger::init_from_env(env_logger::Env::default().default_filter_or("info"));
 
-    let last_sync = web::Data::new(Mutex::new(None));
-    let cloned_last_sync = last_sync.clone();
     actix_rt::spawn(async move {
-        scraping::continous_scraping::start_scraping(cloned_last_sync).await;
+        scraping::continous_scraping::start_scraping(
+            Duration::from_secs(SECONDS_PER_DAY),
+            chrono::Duration::days(30 * 4),
```
octycs, line 31: Maybe it's better to also create a constant here (like `EVENTS_TIME_WINDOW_IN_DAYS`, does that make sense?) to make clear what the numbers mean. I understand it that the calendar API provides all events from 2 months before until 2 months after now.

server/calendar/src/scraping/continous_scraping.rs
```diff
@@ -3,32 +3,25 @@ use crate::scraping::task::ScrapeRoomTask;
...
 use futures::future::join_all;
 use log::{info, warn};
 use std::time::{Duration, Instant};
-use tokio::sync::Mutex;
 use tokio::time::sleep;
 
-const SECONDS_PER_DAY: u64 = 60 * 60 * 24;
-pub async fn start_scraping(last_sync: Data<Mutex<Option<NaiveDateTime>>>) {
-    let mut interval = actix_rt::time::interval(Duration::from_secs(SECONDS_PER_DAY)); //24h
+pub async fn start_scraping(scrape_every: Duration, duration: chrono::Duration) {
```
octycs, line 14:
```suggestion
pub async fn start_scraping(scrape_every: Duration, time_window: chrono::Duration) {
```
Here and further below I think this is more clear than just duration (I was first confused)

octycs (approved):
Looks good, can't find any ~major~ issues. I was just first confused about what "duration" you meant, so maybe timewindow / timeframe is a better name

server/calendar/src/main.rs
CommanderStorm: I think this would not be the best design choice.
Eventually, I want to have more than these 4 Months, but scraped smarter than front to back. (this is a different PR)
Meanwhile, I am working on a follow-up PR, which adds another scraper (longer `duration` and`scrape_every` with lower request rate)

octycs: Okay yes if it's just temporary I can understand

octycs:
From my point it's ready to be merged

PR #430, opened 2023-03-06
Split coordinates file into multiple files

octycs:
The coordinate list has become quite long with the MW and now CH rooms in it, so this PR splits out these larger parts of the list into separate files.

- Split the `02_coordinates.yaml` file into multiple yaml files in a `coordinates` directory

## How to test this PR

1. Recompile the data
2. Verify that coordinates are merged from all the files.

## How has this been tested?

- Verified that coordinates are merged in

## Checklist:

- [x] I have run the linter


data/processors/merge.py
```diff
@@ -43,12 +44,14 @@ def add_translatable_str(value: Union[str, list[Any], dict[str, Any]]) -> Any:
...
 def add_coordinates(data, path):
     """
-    Merge coordinates from the yaml file at path on top of the given data.
+    Merge coordinates from yaml files placed at path on top of the given data.
+    (Merging happens in alphanumeric order, so later files would overwrite earlier files)
     This operates on the data dict directly without creating a copy.
     """
-    yaml_data = load_yaml(path)
+    for fname in os.listdir(path):
+        yaml_data = load_yaml(os.path.join(path, fname))
```
CommanderStorm, line 52: I like the `pathlib.Path` syntax of iterating through files a bit more, as this way this becomes more readbale:
```suggestion
    for file in Path(path).iterdir():
        yaml_data = load_yaml(file)
```
(and either extracting the path from the file or modifing/looking at `load_yaml`)

CommanderStorm (approved):
This is mostly a good PR (I have one small code-style comment, but you can overrule it)

Can be merged if you want

Issue #440, opened 2023-03-12
[Feature] Add Exported Metrics


CommanderStorm:
**Is your feature request related to a problem? Please describe.**
Currently all our (prometheus) metrics come from outside of the system.
While this is mostly okay, it would probably be better if we had internal metrics too. 

**Describe the solution you'd like**
I would like us to export metrics about:

calendar
- [x] how long each scraping run took + when it happened
- [x] how many days are being scraped
- [x] how successfull it was
- [x] how many events are being returned

main API:
- [x] how often each detail (`/api/get/`) has been requested (2 Categories: Bot Traffic/Human Traffic)
- [x] how often each detail (`/api/legacy_redirect/`) has been requested
- [x] <del>how long each of the subqueries in getting the search result took</del> (should be solved via traces)
- [x] how long geting map images for the preview API took
- [ ] how many cache hits there were (how many had to be retrieved)

**Describe alternatives you've considered**
-

**Additional context**
See https://grafana.nav.tum.sexy

CommanderStorm:
Cache hits can be approximated pretty well via the time a request took (anything <1ms is a cache hit 😉)
imo not needed anymore => closing

PR #445, opened 2023-03-14
[metrics-1] added basic metrics

CommanderStorm:
Adds these mertics:
- how often each detail (`/api/get/`) has been requested (2 Categories: Bot Traffic/Human Traffic)
- how often each detail (`/api/legacy_redirect/`) has been requested
- how long geting map images for the preview API took

From #440 

added the following metrics:
- `navigatum_APPNAME_incoming_requests` (labels: endpoint, method, status): the total number of HTTP requests handled by the actix HttpServer.
- `navigatum_APPNAME_response_code` (labels: endpoint, method, statuscode, type): Response codes of all HTTP requests handled by the actix HttpServer.
- `navigatum_APPNAME_response_time` (labels: endpoint, method, status): Total the request duration of all HTTP requests handled by the actix HttpServer.

## How to test this PR

1. Launch one of the servers
2. Do some requests
3. ```bash
   curl http://localhost:8080/metrics
   ```
   (note that for example a browser does not work, as these are not meant to be consumed by a browser)

## How has this been tested?

- curl request

## Checklist:

- [x] I have run the linter


CommanderStorm:
This opens us up to a trivial DOS possiblity ⇒ is thus unaceptable
See https://github.com/nlopes/actix-web-prom/pull/46 for further details

This gives me the motivation needed to look if migrating to rocket and https://github.com/sd2k/rocket_prometheus is a better plan of action

CommanderStorm:
I have checked out [rocket](https://rocket.rs/) and the features.
While they are really nice and some of the notation is way better, this comes at a cost:
This notation is achieved via procedural macros, which don't work sooo great with my IDE ⇒ not an option.

I have checked if this is really exploitable and have come to this conclusion:
While this is definitely exploitable, the effect is very limited.
The main reason is, that we filter URLs in our central ingress ⇒ employability is limited.

I have looked into, if we can provide a patch to [nlopes/actix-web-prom](https://github.com/nlopes/actix-web-prom) or [atomix-team/actix-web-prometheus](https://github.com/atomix-team/actix-web-prometheus): This is not possible due to the architecture of actix (what routes are available is not exposed to middlewares as far as I can tell)

Furthermore, the impact of this is, that Prometheus gets taken offline ⇒ this is picked up by Grafana ⇒ we are alerted and could fix this
Impact is only limited to one of our supporting services ⇒ not production

Thus, while not being ideal, I think this is a barely acceptabe risk (I am obviously not happy, but this will work)

server/calendar/src/main.rs
```diff
@@ -29,6 +31,18 @@ async fn main() -> std::io::Result<()> {
     actix_rt::spawn(async move {
         scraping::continous_scraping::start_scraping(cloned_last_sync).await;
     });
+
+    // metrics
+    let labels = HashMap::from([(
+        "revision".to_string(),
+        std::env::var("GIT_COMMIT_SHA").unwrap_or("development".to_string()),
+    )]);
+    let prometheus = PrometheusMetricsBuilder::new("naviatum_calendar")
```
octycs, line 40:
```suggestion
    let prometheus = PrometheusMetricsBuilder::new("navigatum_calendar")
```

server/main-api/src/main.rs
```diff
@@ -24,6 +25,16 @@ async fn health_status_handler() -> HttpResponse {
 #[actix_web::main]
 async fn main() -> std::io::Result<()> {
     env_logger::init_from_env(env_logger::Env::default().default_filter_or("info"));
+    // metrics
+    let labels = HashMap::from([(
+        "revision".to_string(),
+        std::env::var("GIT_COMMIT_SHA").unwrap_or("development".to_string()),
+    )]);
+    let prometheus = PrometheusMetricsBuilder::new("naviatum_mainapi")
```
octycs, line 33:
```suggestion
    let prometheus = PrometheusMetricsBuilder::new("navigatum_mainapi")
```

server/feedback/src/main.rs
```diff
@@ -53,6 +55,17 @@ async fn main() -> std::io::Result<()> {
         .finish()
         .unwrap();
 
+    // metrics
+    let labels = HashMap::from([(
+        "revision".to_string(),
+        std::env::var("GIT_COMMIT_SHA").unwrap_or("development".to_string()),
+    )]);
+    let prometheus = PrometheusMetricsBuilder::new("naviatum_feedback")
```
octycs, line 63:
```suggestion
    let prometheus = PrometheusMetricsBuilder::new("navigatum_feedback")
```

octycs (approved):
Just a typo but apart from that seems all good. If I understand [this](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) correctly, including the `Cargo.lock` files is better. I was not sure about that.

Since we do not store permanent data in any service that cannot be reretrived, I only see a risk of downtime but no data loss in the case someone tries to spam our metrics until storage space runs out.

PR #452, opened 2023-03-18
[Cal-7] Cron based calendar scraping

CommanderStorm:
Fixes a conceptual bug in how the calendar scrapes:

Previously, we were scraping 
- at constant time intervals (as in: not recognizing that we should only scrape at night)
- only with one scrape-lenght
- in production AND in staging environments (⚠️)

- Moved the code for the scraper into its binary (these look intimidating, but are just code moves)
- Changed the deployment of the calendar back to being in a `Deployment`, but one with a `PersitantVolume` attached
- Added `ChronJobs` which control when this functionality is run

## How to test this PR

1.  Request this API 
2. Code review (The big additions/deletions come from code moves)

## How has this been tested?

- [x] Smoke tests of the deployment

## Checklist:

- [x] I have run the linter


server/calendar/src/models.rs
```diff
@@ -0,0 +1,29 @@
+use chrono::NaiveDateTime;
+use diesel::prelude::*;
+use diesel::Insertable;
+
+#[derive(Insertable, Queryable, AsChangeset)]
+#[diesel(table_name = crate::schema::calendar)]
+pub struct XMLEvent {
```
CommanderStorm, line 7: This had to be in its module, as this is used in the scraper and regular api

deployment/k3s/templates/deployments/backend/calendar-cronjob.yaml
```diff
@@ -0,0 +1,42 @@
+{{ range $scrape_task := $.Values.calendar.scrape_tasks }}
```
CommanderStorm, line 1: before merging, we have to add a feature gate for this, to only be available in master.
This is currently not done, because Otherwise this would be untestable

octycs (approved):
I don't have a lot of experience about the K3S part (so I'm not sure what you mean with your comment on that file or if it still needs to be fixed). But the Rust part looks all fine to me :+1:

PR #453, opened 2023-03-18
[Cal-8] `reqwest` based calendar scraping

CommanderStorm:
In #452 I noticed, that the API of `awc` can be simplified with slightly a less performant library.

Reasoning:
- Performance for the scraper is not an issue, as we scrape with a scrape-budget anyway
- `awc` has requirements, from the tokio environment, which would look very ugly => imo this is way better

- migrated from awc to reqwest

## How to test this PR

1. Code review

## How has this been tested?

- Ran the code

## Checklist:

- [x] I have run the linter

### DIFF
--- a/server/calendar/src/scrape_task/tumonline_calendar_connector.rs
+++ b/server/calendar/src/scrape_task/tumonline_calendar_connector.rs
@@ -1,8 +1,6 @@
 use crate::models::XMLEvent;
 use crate::scrape_task::scrape_room_task::ScrapeRoomTask;
 use crate::{schema, utils};
-use awc::error::{ConnectError, PayloadError, SendRequestError};
-use awc::Client;
 use chrono::{NaiveDateTime, Utc};
 use diesel::prelude::*;
 use log::{debug, error, warn};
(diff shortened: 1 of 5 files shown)


server/calendar/src/scrape_task/tumonline_calendar_connector.rs
```diff
@@ -16,61 +14,39 @@ enum RequestStatus {
...
-async fn request_body(client: &Client, url: String) -> RequestStatus {
-    let req = client.get(&url).send().await;
+
+async fn request_body(url: String) -> RequestStatus {
+    let req = reqwest::get(&url).await;
     let body = match req {
-        Ok(mut res) => match res.status().as_u16() {
-            200 => res.body().limit(2_usize.pow(32)).await,
+        Ok(res) => match res.status().as_u16() {
+            200 => res.text().await,
```
octycs, line 25: Is the limit (`res.body().limit(2_usize.pow(32))` beforehand) no longer required for reqwest or just no longer in general (bc 2³² is a lot anyway)?

octycs (approved):

server/calendar/src/scrape_task/tumonline_calendar_connector.rs
CommanderStorm: `reqwest` has a different API for specifying if a limit is required.
We want to scrape indiscriminately (requests are smaller than memory => not an issue)

see https://www.reddit.com/r/rust/comments/mkk1eb/comment/gtg9jox/

[Streaming the text](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.text) could be a oppertunity to save this memory, but would require a different processing approach.
The current approach works => I dont think we need to improve this

Issue #424, opened 2023-03-02
[Feature] Automatic regular TUMonline updates


octycs:
**Is your feature request related to a problem? Please describe.**
TUMonline seems to be updated quite a lot (more than I thought) and rooms appear to change more often than every semester.
To keep our data up to date, we should automate our TUMonline updates

**Describe the solution you'd like**
The best idea I have so far is to use GitHub actions [schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) feature to update e.g. weekly our TUMonline base data by creating a PR.

**Describe alternatives you've considered**
The above is just an idea, open for discussion.

**Additional context**
Instead of the tedious current scraper, it would make sense to switch to the TUMonline XML API that we have been provided (with a token).

Belongs somewhat to #197

CommanderStorm:
Resolved by https://github.com/TUM-Dev/NavigaTUM/pull/1114

PR #456, opened 2023-03-19
[autoscraper-1] merged `_write_cache_json` and `_cached_json` into a single decorator

CommanderStorm:
In the PR-Train for #424 I noticed that the caching feature would be better served by having a more ergonomic caching function

- added a decorator `@cached_json(filename)` which can be added to functions
- said decorator supports formatting

## How to test this PR

```python
@cached_json("abc")
def t():
    return []


@cached_json("abc_{b}")
def t_1(a, b2, b=1, cd=2):
    return []


if __name__ == "__main__":
    t_1(1, 2, b=3)
```

## How has this been tested?

- see above

## Checklist:

- [x] I have run the linter

### DIFF
--- a/data/external/scrapers/nat.py
+++ b/data/external/scrapers/nat.py
@@ -3,7 +3,7 @@ import json
 import logging
 from multiprocessing.pool import ThreadPool
 
-from external.scraping_utils import _cached_json, _download_file, _write_cache_json, CACHE_PATH
+from external.scraping_utils import _download_file, CACHE_PATH, cached_json
 from tqdm import tqdm
 from tqdm.contrib.concurrent import thread_map
 from utils import TranslatableStr as _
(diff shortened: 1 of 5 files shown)


data/external/scrapers/nat.py
```diff
@@ -171,10 +158,9 @@ def _get_base_room_infos():
             work_queue = new_queue
 
     total_hits = _join_room_hits()
-    _write_cache_json(cache_name, total_hits)
-    if undownloadable:  # down here to make shure, that tdtm has flushed the output
+    if undownloadable:  # down here to make sure, that tdtm has flushed the output
         _report_undownloadable(undownloadable)
-    return total_hits
+    return total_hits  # noqa: R504
```
octycs, line 163: What does this mean?

octycs (approved):
Looks all good to me

data/external/scrapers/nat.py
CommanderStorm: Thsis is here to tell flake8, that the warning `unnecessary variable assignment before return statement` should be ignored

PR #471, opened 2023-03-24
[Cal-11] Postgres

CommanderStorm:
Currently we use SQlite to store our data.
This approach has a few downsides:
- SQLite manages state in one file (i.e. writes lock the entire database)
- Since all Staging branches share this state, they need to be scheduled on the same node, making the system unnecessarily brittle

This PR migrates this datastore to Postgres, which should be a better long-term solution to this problem

- Migration from sqlite to postgres as a datastore for the Calendar

## How to test this PR

- code review

## How has this been tested?

1. Launch `docker run -p 5432:5432 -e POSTGRES_PASSWORD=password postgres`
2. Lanch the calendar
3. See it continuing to scrape

## Checklist:

- [x] I have run the linter


deployment/k3s/templates/deployments/calendar/postgres-volume.yaml
```diff
@@ -1,29 +1,29 @@
...
     app: navigatum
-    deployment: maps
+    deployment: postgres
     type: local
 spec:
   storageClassName: local-path
   capacity:
     storage: 1Gi
-  persistentVolumeReclaimPolicy: Retain
+  persistentVolumeReclaimPolicy: Delete
```
octycs, line 13: Does this mean a redeployment of Postgres removes all the data?

server/calendar/README.md
```diff
@@ -13,12 +13,14 @@ Please follow the [system dependencys docs](resources/documentation/Dependencys.
...
 
-To set up the database, you will need to run the `init_db.py` script:
+To set up the database, you will need to run a `postgres` instance. We recommend this configuration:
 
 ```bash
-python3 init_db.py
+docker run -p 5432:5432 -e POSTGRES_PASSWORD=password postgres
 ```
 
+Any migrations this database needs are applied on first run of the server.
```
octycs, line 22:
```suggestion
Any migrations this database needs are applied on first run of the server.
In order to change the connection parameters that the calendar server uses you can use the environment variables:
- `POSTGRES_USER` (default `postgres`)
- `POSTGRES_PASSWORD` (default `password`)
- `POSTGRES_URL` (default `localhost`, specify the port if different from 5432)
- `POSTGRES_DB` (default: same as `POSTGRES_USER`)
```

octycs (approved):
Yes Postgres is probably a better long term solution for our data. The migrations thing is nice.

deployment/k3s/templates/deployments/calendar/postgres-volume.yaml
CommanderStorm: No, if a volume would be deleted (i.e. manually), this means this volume is not Deleted instead of Retained. Given that this data is relatively eazy to recover, I think `Delete` is probably the better choice

Issue #501, opened 2023-04-09
[Feature] Better Feedback Pipeline


CommanderStorm:
**Is your feature request related to a problem? Please describe.**
The current pipeline for accepting coordinate/image/… changes is very labor-intensive.

Current workflow:
- User creates issue(s)
- We have to manually go into the code, add the change and sanity check if this makes sense
- We have to close the issue

**Describe the solution you'd like**
I would like this kind of feedback as a PR instead

Work Plan to get this done:
- [x] Refactor the Feedback system to improve the code-adaptability
- [x] change the current Feedback-API (renaming `delete_issue_requested` to `deletion_requested`)
- [ ] implement an API which can receive edit-requests
- [ ] add this change to the frontend (dependent on #200)
- [ ] Do some user Testing on the new workflow

**Describe alternatives you've considered**
I don't think there are better alternatives out there (at least I can't think of any)

PR #502, opened 2023-04-09
[feedback-pipeline-1] refactored the feedback system to be flatter structured

CommanderStorm:
First PR in the #501 PR-Train

- Refactored how the previeous feedback system is laid out (particularly in terms of where files are located)

## How to test this PR

1. Code review (check if you agree with how I moved code around)

## How has this been tested?

- Code review
- Smoke/unit tests

## Checklist:

- [x] I have run the linter


server/feedback/src/core.rs
```diff
@@ -0,0 +1,53 @@
...
+        self.feedback_keys.github_token.is_some() && self.feedback_keys.jwt_key.is_some()
+    }
+}
+
+pub struct TokenRecord {
+    pub kid: u64,
+    pub next_reset: usize,
+}
+
+pub async fn get_token(state: Data<AppStateFeedback>) -> HttpResponse {
```
octycs, line 30: Would it make sense to move `get_token` into `tokens.rs`?

CommanderStorm: True.
I have no idea why I came up with this architecture in the first place.
I have refactored this to live in the token directory and simplified the acompaniing code

PR #507, opened 2023-04-10
[feedback-pipeline-4] `image-sources.yaml` regularisation

CommanderStorm:
PR in the https://github.com/TUM-Dev/NavigaTUM/issues/501 PR-Train

Main motivation:
The further PR train requires `image-sources.yaml` to be regularly formatted.

- Removed the ability for `image.author`s to have `url`s:
  The reasoning is that this only occurs on two fields and having `OR-Types` does not serialize particularly well.
  Upstream API is kept, as this would be a breaking change which probably impacts other usages.
- Removal of the licence`-shorthand:
  As above, this is an `OR`-Type ⇒ not easy to support
- Removal of `meta`: Honestly, I don't see the point why we could ever integrate this. This data is just too scattered.
- Reformatting: The rust library reformats quite a few things.
   Strings now have no quotes if they don't have to
   Sadly, no rust library currently supports keeping comments ⇒ They are no longer present.
   

## How to test this PR

1. Check if the changes I made are Okay (Go through the commits, don't look at the full summary 😉)

## How has this been tested?

- Code review

## Checklist:

- [x] I have run the linter

### DIFF
--- a/data/processors/images.py
+++ b/data/processors/images.py
@@ -101,20 +101,12 @@ def _add_source_info(fname, source_data):
             return {"text": obj, "url": None}
         return obj
 
-    img_data = {
+    return {
         "name": fname,
         "author": _parse(source_data[_index]["author"]),
         "source": _parse(source_data[_index]["source"]),
+        "license": _parse(source_data[_index]["license"]),
     }
-    # add license information utilising shorthands if available
-    raw_license = source_data[_index]["license"]
-    if isinstance(raw_license, str):
-        if raw_license in KNOWN_LICENSE_URLS:
-            img_data["license"] = {"text": raw_license, "url": KNOWN_LICENSE_URLS[raw_license]}
-            return img_data
-        logging.warning(f"Unknown license url for licence shorthand '{raw_license}'. No url will be added")
-    img_data["license"] = _parse(raw_license)
-    return img_data
 
 
 def _gen_fixed_size(img: Image.Image, fixed_size: tuple[int, int], offset: int) -> Image.Image:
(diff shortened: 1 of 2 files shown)


octycs:
> Removal of meta: Honestly, I don't see the point why we could ever integrate this. This data is just too scattered.

I would at least prefer to keep the `date` field (even though it is not uniformly formatted at the moment, that should be a needed fix) since it would help to indicate if an image is outdated. It could be null if we don't know it.

CommanderStorm:
> If it's removed, also https://github.com/TUM-Dev/NavigaTUM/blob/main/data/processors/images.py#L29 would need to be removed

Done

> I would at least prefer to keep the `date` field (even though it is not uniformly formatted at the moment, that should be a needed fix) since it would help to indicate if an image is outdated. It could be null if we don't know it. Possibly also `caption` that can serve as image `alt` attribute.

I have kept the `meta` attributes

octycs:
> I have kept the meta attributes

Maybe we can standardize that there are only `date` and `caption`, but no other fields? I don't want to neglect your opinion on that.

CommanderStorm:
> Maybe we can standardize that there are only `date` and `caption`, but no other fields? I don't want to neglect your opinion on that.

I would argue that this would need additional support (i.e. changing the API/Frontend) and need an additional Review, given that a lot of these data points would have to be reworded (some of they contain a ton of text which should not be in a caption)
⇒ Your Point is better, let's keep them and integrate this down the line.
They currently don't hurt us badly 😉

This was not a strong opinion about me never seing us integrating it…

octycs (approved):

PR #509, opened 2023-04-10
[POI-1] Add basic POI support in the data

octycs:
Relates to #499 

- Add support for data entries of type `poi` intended for entries with a location that are not rooms/buildings etc.
- Includes "Validierungsautomat 1" as example, the rest still needs to be added (coordinates researched)

## How to test this PR

1. Recompile the data
2. Open the website and go to `/view/parabelrutsche`. Search is not yet supported in this PR, only external linking.

## How has this been tested?

- "Parabelrutsche" and "Validierungsautomat 1" are displayed correctly

## Checklist:

- [x] I have run the linter


data/sources/21_pois.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+  props:
+    generic:
+      - name:
+          de: "Formel"
+          en: "Formula"
+        text: "z = y = hx²/d²"
+
+# === Validierungsautomaten ===
+validierungsautomat-1:
```
CommanderStorm, line 17: Could we add all the different machines:
The [documentation for which building they are](https://www.tum.de/studium/bewerbung/infoportal-einschreibung/studentcard) in is clear enough for most, exact coordinates could then be added via the normal feedback..

data/processors/structure.py
```diff
@@ -104,6 +104,7 @@ def _get_type(_id, _data):
             "building": _("Gebäude"),
             "room": _("Raum"),
             "virtual_room": _("Raum/Gebäudeteil"),
+            "poi": _("POI"),
```
CommanderStorm, line 107: I have not tested this, but does this work (given that `en`=`de` in this case):
```suggestion
            "poi": "POI",
```

data/processors/poi.py
```diff
@@ -0,0 +1,27 @@
...
+
+        parent = data[poi["parent"]]
+        poi["parents"] = parent["parents"] + [parent["id"]]
+        del poi["parent"]
+
+        poi.setdefault("sources", {}).setdefault("base", []).append(
+            {
+                "name": "NavigaTUM",
+            },
+        )
```
CommanderStorm, line 25: Your point here is probably, that we did some work and should be in the sources?
I would argue that if somebody added this while explicitly specifying the `source`, adding ourselves would not be "correct" (not a strong opinion)

```suggestion
        poi.setdefault("sources", {"base": [{"name": "NavigaTUM"}]})
```

data/processors/poi.py
```diff
@@ -0,0 +1,27 @@
...
+            raise ValueError(f"The id '{_id}' is already used, cannot use it for a new POI")
+
+        if poi["parent"] not in data:
+            raise ValueError(f"Parent '{poi['parent']}' of POI '{_id}' not found")
+
+        poi["type"] = "poi"
+
+        parent = data[poi["parent"]]
+        poi["parents"] = parent["parents"] + [parent["id"]]
+        del poi["parent"]
```
CommanderStorm, line 19:
```suggestion
        parent_id = poi.pop("parent")
        parent = data[parent_id]
        poi["parents"] = parent["parents"] + [parent["id"]]
```

data/translations.yaml
```diff
@@ -42,6 +42,7 @@ Gebäudekomplex: Building complex
 Gebäude: Building
 Raum: Room
 Raum/Gebäudeteil: Room/Part of the building
+POI: POI
```
CommanderStorm, line 45: see above

data/sources/21_pois.yaml
```diff
@@ -0,0 +1,28 @@
...
+  parent: "5101"
+  name: "Validierungsautomat 1 (Physik)"
+  coords: { lat: 48.267097, lon: 11.675607 }
+
+  usage:
+    name: "Validierungsautomat"
```
CommanderStorm, line 23: How should these things be internationalized?
I think we could run `usage` through our `_(...)` translation function?

CommanderStorm (changes requested):
The only real issues I noticed, are:

the sitemap probably should be robuster 😉 
image

I think this should also be added here:
https://github.com/TUM-Dev/NavigaTUM/blob/main/webclient/src/views/view/view-view.js#L26-L35

data/processors/poi.py
octycs: > Your point here is probably, that we did some work and should be in the sources?

I didn't pay enough attention to the code i copied over and adapted. You're right of course, it should be like this.

octycs:
> I think POI should also be added here:
https://github.com/TUM-Dev/NavigaTUM/blob/main/webclient/src/views/view/view-view.js#L26-L35

Yes, this should be adapted. I wanted to avoid changes webclient in this PR because of #200 and because it's not strictly necessary for POIs to work in the webclient (`/poi/<id>` instead of `/view/<id>` is just to make URLs nicer but doesn't break anything).

data/sources/21_pois.yaml
```diff
@@ -0,0 +1,28 @@
+parabelrutsche:
+  parent: "mi"
+  name: "Parabelrutsche"
+  coords: { lat: 48.26254, lon: 11.66796 }
```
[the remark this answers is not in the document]
CommanderStorm: One advantage of adding them to the `coords` is that they won't require special attention in #501, but this also would not be a breaking issue for #501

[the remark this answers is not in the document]
octycs: I moved them over. I also added that they can inherit their coordinate from the parent if it is not known. This way we can at least indicate the building for validation machines where we don't know the exact location.

data/sources/21_pois.yaml
octycs: I have added all machines and their coordinates if I could find them out (if not there's sometimes a coment; for the MI thanks to [the Campus Tour](https://youtu.be/rQNNL1tvmDQ?t=10) ^^ ). For two machines I couldn't find out their number from the [HTML source of the status page](https://campus.tum.de/valistatus/) like for the others:
- TUM Campus Straubing Bibliothek
- TUM School of Education

octycs:
Given the problem with the sitemap I've now also added the URLs to the webclient

data/sources/21_pois.yaml
CommanderStorm: The CS is here: https://nav.tum.de/building/2927
The "Edu" one is here: https://nav.tum.de/room/2907.01.130

Source: https://www.tum.de/studium/bewerbung/infoportal-einschreibung/studentcard
(Section Standorte der Validierungsstationen)

octycs: Oh sorry I was not clear maybe. I couldn't find out their number like "Station Nr. 42", so their IDs are currently `validierungsautomat-straubing` and not `validierungsautomat-42`

CommanderStorm (approved):
This can be merged now imo
There is a styling comment open, which I would like to see merged, but this is something I can live without.
Note that during #508 the coordinates will have to be integrated into `rest.yaml`

PR #561, opened 2023-05-03
Add floor patch for 0507.03.750

### LINKED ISSUE #560: [Entry] [0507.03.750]:
Das Stockwerk in diesem flügel ist die erste Zahl der Architektennummer  
=> Das hier ist das 2. Stockwerk  
~Frank

octycs:
Resolves #560

(as PR so we can discuss)

- Patch the floor of this lecture hall to 2
- Add an informational comment like for [MW1801](https://nav.tum.de/room/5508.02.801). If you consider it unnecessary we can remove the comment though

## How to test this PR

1. Open 0507.03.750 in a patched instance

## How has this been tested?

- 

## Checklist:

- [ ] I have run the linter

### DIFF
--- a/data/sources/02_rooms-extended.yaml
+++ b/data/sources/02_rooms-extended.yaml
@@ -8,6 +8,14 @@
   comment:
     de: Unter dieser Treppe existiert ein WC, dessen Raumnummer und designierung (men/women/unisex) unklar ist
     en: Under this staircase there is a toilet whose room number and designation (men/women/unisex) is unclear
+"0507.03.750":  # Karl Max von Bauernfeind Hörsaal
+  props:
+    comment:
+      de: Dieser Hörsaal ist in TUMonline 0507.03.750 statt 0507.02.750, obwohl der Hörsaal im 2. Stock ist. Aus Konsistenzgründen übernehmen wir diese ID.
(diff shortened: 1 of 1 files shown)


CommanderStorm:
I think that this is a more general issue (we don't handle that lecture halls normally span two floors)

Adding a comment for every one is imo too labour-intensive, given that we have about 130 Lecture halls (source: this is the number ProLehre quoted us for Livestreaming)

I dont know what is the best solution until we can detect this via the CAD maps from #403 (still ongoing, progress is slower than expected, sorry..)

CommanderStorm:
We can merge this or close the underlying issue (I am indifferent about which one we choose)

I think that the comment should be `This {display_name} has entries on {floor1} and {floor2}`. until we support rooms spanning floors...

octycs:
I agree, lecture halls (or any other rooms) with entries in more than floor cannot be handled at the moment, and I'm also not aware of any way to detect that from our data (e.g. the interims I lecture halls also don't have a clear main entrance). Until there is a solution I would stick to using the main entrance if there is one and else whatever TUMonline uses, so we have fewer patches. I am not sure if this particular lecture hall has a clear main entrance

PR #598, opened 2023-05-13
[Search-Refactoring-2] `reqwest` based main api

CommanderStorm:
This PR is similar in content as #453
In https://github.com/TUM-Dev/NavigaTUM/pull/452 I noticed, that the API of awc can be simplified with slightly a less performant library.

- Performance for the scraper is not an issue, as we scrape with a scrape-budget anyway
- `awc` has requirements, from the tokio environment, which would look very ugly => imo this is way better

- migrated from awc to reqwest

## How to test this PR

1. Code review

## How has this been tested?

- Ran the code

## Checklist:

- [x] I have run the linter

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1540,7 +1540,7 @@ components:
         time_ms:
           description: |
             Time the search took in the server side, not including network delay
-            Maximum as by awc timeout. other timeouts (browser, your client) may be smaller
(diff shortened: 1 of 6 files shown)


CommanderStorm:
merging as not critical to the mission and a similar PR has already been acccepted

PR #599, opened 2023-05-14
[Search-Refactoring-3] Lexer/parser based searching

CommanderStorm:
Switched from a custom preprocessor to a more regular lexer/parser architecture
The main motivation is that the current code is quite unmaintainable. Adding features like #106 in the current state may lead to disastrous results.

- Replaced the custom preprocessor with [`logos`](https://docs.rs/logos/latest/logos/)
logos

## How to test this PR

1. Code review

## How has this been tested?

- [x] Unittests of the Lexer
- [x] Unittests of the Parser

## Checklist:

- [x] I have run the linter


Fizan324926 (commented):
### Logging
The code includes a logging statement using log::warn in the From<&str> for ParsedQuery implementation in parser.rs. It’s generally a good practice to provide meaningful error messages or logs to aid in debugging and troubleshooting. Make sure that the actual error message is printed by using string interpolation correctly, such as warn!(”Error in query parsing: :?”, e) instead of warn!(”Error in query parsing: e:?”)

### Use of Cloning
In the query.rs file, the use of clone() on the tokens field in the prompt_for_queriing and prompt_for_queriing_room methods may introduce unnecessary overhead if the tokens field contains a large amount of data. Consider whether cloning the tokens is necessary or if there are alternative approaches that avoid cloning.
### Unit Tests
I commend the inclusion of unit tests for the lexer and parser. The tests cover various scenarios and help ensure the correctness of the implementation. It’s great to see the test coverage expanding.
**Suggestion** Consider expanding the test coverage to include additional edge cases or scenarios specific to the requirements of the search functionality.

### Code Formatting and Documentation
The code adheres to Rust’s formatting guidelines, and the naming conventions are consistent and meaningful. The newly added files follow the module structure and naming conventions used throughout the project.

### Deprecated Function
In the parser.rs file, the len function in the ParsedQuery struct is marked as deprecated with a note explaining its purpose as a workaround. Deprecated code indicates that there might be a better or more conventional approach to achieve the same result. It’s worth considering alternative solutions to address the underlying issue.

server/main-api/src/search/search_executor/lexer.rs
```diff
@@ -0,0 +1,167 @@
...
+pub enum Token {
+    #[regex("\"[^\"]+\"", | lex | lex.slice()[1..lex.slice().len() - 1].to_string(), priority = 3)]
+    #[regex(r"[^ \t\n\f]+", | lex | lex.slice().to_string())]
+    Text(String),
+
+    #[regex("[a-zA-Z]+[0-9]{1,4}", irregular_split, priority = 2)]
+    SplittableText((String, String)),
+
+    #[regex("in: ?[a-zA-Z0-9]+", | lex | remove_prefix(lex, "in:"), priority = 3)]
+    #[regex("@ ?[a-zA-Z0-9]+", | lex | remove_prefix(lex, "@"), priority = 3)]
```
octycs, line 39:
```suggestion
    #[regex("in: ?[a-zA-Z0-9-]+", | lex | remove_prefix(lex, "in:"), priority = 3)]
    #[regex("@ ?[a-zA-Z0-9-]+", | lex | remove_prefix(lex, "@"), priority = 3)]
```
Some IDs have a `-` in them. Is there a reason you added the ` ?`? I don't think it's a problem (except in edge cases when you are about to prepend a filter to a query), but I also wouldn't expect it

server/main-api/src/search/search_executor/lexer.rs
```diff
@@ -0,0 +1,167 @@
...
+    ParentFilter(String),
+
+    #[regex("usage: ?[a-zA-Z0-9]+", | lex | remove_prefix(lex, "usage:"), priority = 3)]
+    #[regex("nutzung: ?[a-zA-Z0-9]+", | lex | remove_prefix(lex, "nutzung:"), priority = 3)]
+    #[regex("= ?[a-zA-Z0-9]+", | lex | remove_prefix(lex, "="), priority = 3)]
+    UsageFilter(String),
+
+    #[regex("type: ?[a-zA-Z0-9]+", | lex | remove_prefix(lex, "type:"), priority = 3)]
```
octycs, line 47:
```suggestion
    #[regex("usage: ?[a-zA-Z]+", | lex | remove_prefix(lex, "usage:"), priority = 3)]
    #[regex("nutzung: ?[a-zA-Z]+", | lex | remove_prefix(lex, "nutzung:"), priority = 3)]
    #[regex("= ?[a-zA-Z]+", | lex | remove_prefix(lex, "="), priority = 3)]
    UsageFilter(String),

    #[regex("type: ?[a-zA-Z]+", | lex | remove_prefix(lex, "type:"), priority = 3)]
```
Probably we don't need numbers there

server/main-api/src/search/search_executor/lexer.rs
```diff
@@ -0,0 +1,167 @@
...
+        let mut lexer = Token::lexer("\" \"\"");
+        assert_eq!(lexer.next(), Some(Ok(Token::Text(" ".to_string()))));
+        assert_eq!(lexer.next(), Some(Ok(Token::Text("\"".to_string()))));
+        assert_eq!(lexer.next(), None);
+        for text in ["a", "a ", "a a ", " a a ", " @ = in: contains: type: a "] {
+            let quoted_text = format!("\"{text}\"");
+            let mut lexer = Token::lexer(&quoted_text);
+            assert_eq!(lexer.next(), Some(Ok(Token::Text(text.to_string()))));
+            assert_eq!(lexer.next(), None);
+        }
```
octycs, line 82: Do you know what happens for the special cases `in:"`, `@"Garching"`, `usage: "lecture hall"` etc? I think allowing quoting for filter values would be a valuable feature, but currently they have the same priority
(I can test this weekend maybe)

server/main-api/src/search/search_executor/lexer.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    #[test]
+    fn normal_splits() {
+        for text in ["foo foo", "foo\nfoo", "foo  foo"] {
```
octycs, line 87:
```suggestion
        for text in ["foo foo", "foo\nfoo", "foo  foo", " foo foo", "foo foo ", " foo foo "] {
```

server/main-api/src/search/search_executor/query.rs
```diff
@@ -42,7 +42,7 @@ impl GeoEntryQuery {
         }
     }
     pub async fn execute(self) -> Result<MultiSearchResponse<MSHit>, Error> {
-        let q_default = self.parsed_input.to_default_query();
+        let q_default = self.prompt_for_queriing();
```
octycs, line 45:
```suggestion
        let q_default = self.prompt_for_querying();
```

server/main-api/src/search/search_executor/query.rs
```diff
@@ -57,11 +57,36 @@ impl GeoEntryQuery {
             .multi_search()
             .with_search_query(self.merged_query(&entries, &q_default))
             .with_search_query(self.buildings_query(&entries, &q_default))
-            .with_search_query(self.rooms_query(&entries, &self.parsed_input.to_room_query()))
+            .with_search_query(self.rooms_query(&entries, &self.prompt_for_queriing_room()))
```
octycs, line 60:
```suggestion
            .with_search_query(self.rooms_query(&entries, &self.prompt_for_querying_room()))
```

server/main-api/src/search/search_executor/query.rs
```diff
@@ -57,11 +57,36 @@ impl GeoEntryQuery {
...
             .execute::<MSHit>()
             .await
     }
 
+    fn prompt_for_queriing(&self) -> String {
```
octycs, line 65:
```suggestion
    fn prompt_for_querying(&self) -> String {
```

server/main-api/src/search/search_executor/query.rs
```diff
@@ -57,11 +57,36 @@ impl GeoEntryQuery {
...
+            .clone()
+            .into_iter()
+            .map(|s| match s {
+                TextToken::Text(t) => t,
+                TextToken::SplittableText((t1, t2)) => format!("{t1}{t2}"),
+            })
+            .collect::<Vec<String>>()
+            .join(" ")
+    }
+    fn prompt_for_queriing_room(&self) -> String {
```
octycs, line 77:
```suggestion
    fn prompt_for_querying_room(&self) -> String {
```

server/main-api/src/search/search_executor/lexer.rs
CommanderStorm: my assumption is that if you are writing `in: garching` a space can slip in, that is why.
Thinking about it: Lets include `.` here to 😅 
```suggestion
    #[regex("in: ?[a-zA-Z0-9-\.]+", | lex | remove_prefix(lex, "in:"), priority = 3)]
    #[regex("@ ?[a-zA-Z0-9-\.]+", | lex | remove_prefix(lex, "@"), priority = 3)]
```

server/main-api/src/search/search_executor/lexer.rs
```diff
@@ -0,0 +1,167 @@
...
+pub enum Token {
+    #[regex("\"[^\"]+\"", | lex | lex.slice()[1..lex.slice().len() - 1].to_string(), priority = 3)]
+    #[regex(r"[^ \t\n\f]+", | lex | lex.slice().to_string())]
+    Text(String),
+
+    #[regex("[a-zA-Z]+[0-9]{1,4}", irregular_split, priority = 2)]
+    SplittableText((String, String)),
+
+    #[regex("in: ?[a-zA-Z0-9-\.]+", | lex | remove_prefix(lex, "in:"), priority = 3)]
+    #[regex("@ ?[a-zA-Z0-9-\.]+", | lex | remove_prefix(lex, "@"), priority = 3)]
```
CommanderStorm, line 39:
```suggestion
    #[regex("in: ?[a-zA-Z0-9-.]+", | lex | remove_prefix(lex, "in:"), priority = 3)]
    #[regex("@ ?[a-zA-Z0-9-.]+", | lex | remove_prefix(lex, "@"), priority = 3)]
```

server/main-api/src/search/search_executor/lexer.rs
CommanderStorm: Reason for the priority being the the same is, that `"..."` and `in:...` have a different start => can never intersect

I think this is a special case which we currently probably do not need.
I have added a testcase documenting this behaviour.

Examples:
`in:foo` -> `ParentFilter("foo")`
`in: foo` -> `ParentFilter("foo")`
`in:"foo"` -> `Text("in:"foo"")`
`in: "foo` -> `Text("in:")`, `Text(""foo")`
`in: "foo"` -> `Text("in:")`, `Text("foo")`

CommanderStorm: Proposed Solutions:
- if a usage includes spaces, we should encode it differently (e.g. `lecture`_hall`)
- given that the others are also only direct-match filters, this should not be nessesary

CommanderStorm: I am merging this PR, as the problems noticed here are not major enough imo

Issue #86, opened 2022-04-24
[Feature] Show close MVV stations


octycs:
**Is your feature request related to a problem? Please describe.**
While we have an interactive map where a building or site is located, people still have to look somewhere else to see how to get there.

**Describe the solution you'd like**
On site / building pages, we can show a list of close-by public transport (Subway, Bus) stations, so people see how to get there. We could either collect them by hand for each site / building, or maybe even better collect a list of them with coordinates and auto-generate. Then we could even display the distance in meters.

**Describe alternatives you've considered**
We could also just add links to OSM / Google Maps based on the coordinates. But I think this isn't actually mutually exclusive, we could do both.

**Additional context**
Tasks could involve:
- Getting a list of relevant public transport stations (please also credit them in `webclient/src/md/ueber-uns.md` and `webclient/src/md/about-us.md`) 
- Adding a new section specification in `sections:` in `data/data_format_geo-entry.yaml` and implement it in e.g. a new section-generating step after `-- 81 Generate overview sections`
- Implement the display HTML/CSS in `webclient/src/views/view/view-view.inc/.scss`

CommanderStorm:
I am goonna drop this link here:
https://python-mvg-departures.readthedocs.io/en/latest/#mvg_api.get_nearby_stations

Wayrion:
For future reference, there seems to be a newer API which supports the latest MVG v3 Endpoint which more performant and the API wrapper supports async usage: https://github.com/mondbaron/mvg

CommanderStorm:
Forgot to update this issue:

Things in public transit have moved quite a bit and now [DELFFI e.V.](https://www.delfi.de/) is publishing realtime transit data.
That data is pretty close to the ground truth in the few "benchmarks" I did.

The major war-plan for navigatum is to use a mix of routers for the navigation feature. One of them is [motis](https://motis-project.de/).
This router also has the ability to cheaply get a list of nearby stations and cheaply get what the departures at these stations are.
We should use this instead of the mvg to have this also work on the non-mvg campuses like heilbron, straubing, ...

I have not uptdated `/api/locations/{id}/nearby` to use the motis router (the router that we will be using), but will do some engineering into this direction in the next part of the day.

PR #663, opened 2023-06-12
[MVV-2] Show close MVV stations

quarz12:
Implements #86

- [x] use data from [mvv](https://www.mvv-muenchen.de/fahrplanauskunft/fuer-entwickler/opendata/index.html) to generate json file with all stations
- [x] nearby function that filters out all stations that are not within a given distance with binary search
- [ ] display the data on the webpage 

## How to test this PR

test for json data: in test_models.py

## Checklist

- [ ] I have updated the documentation / No need to update the documentation
- [ ] I have run the linter


quarz12:
At this point I have finished the data fetching part of this pr. I used the building coordinates and IDs generated by ```scrape_maps```. I am not really sure if this is the correct way of getting them though. Also, it unfortunatelly looks like the csv provided by the MVV, while extensive, is not complete. 
I will now look into how to display that data. I do not think I will be able to adapt the API though as I have never worked with rust before.

data/external/scrapers/public_transport.py
```diff
@@ -0,0 +1,66 @@
...
+    for building in buildings:
+        # ignore entries that dont have assigned lon/lat
+        if latlonbox := building.get("latlonbox"):
+            # since lat and lon have rather long decimal points it is probably best to not use floats
+            latn, lats = Decimal(latlonbox.get("north")), Decimal(latlonbox.get("south"))
+            lone, lonw = Decimal(latlonbox["east"]), Decimal(latlonbox["west"])
+            lat: Decimal = avg(latn, lats)
+            lon: Decimal = avg(lone, lonw)
+            building_coords.append({"id": building.get("id"), "desc": building.get("desc"), "center-coord": {"lat":str(lat),"lon":str(lon)}, "public-transport":nearby((lat,lon),stations)})
+
```
CommanderStorm, line 65: Conceptually, the scrapers should only scrape the data, which is then merged into the data pipeline.
If necessary, a scraper can rely upon itself, but not other parts.

The benefit of this is that the scrapers are more decoupled and hopefully easier to maintain.

```suggestion
import csv
from decimal import Decimal

from external.scraping_utils import cached_json


@cached_json("public_transport.json")
def scrape_stations():
    with open("scrapers/MVV_HSTReport2212.csv", "r", newline='') as file:
        reader = csv.DictReader(file, delimiter=";")
        return [{
                    "id": int(line["﻿HstNummer"]),
                    "name":line["Name ohne Ort"],
                    "location":line["Ort"],
                    "global_id":line["Globale ID"],
                    "lat":Decimal(line["WGS84 X"].replace(",", ".")),
                    "lon":Decimal(line["WGS84 Y"].replace(",", ".")),
                } for line in reader if line["﻿HstNummer"]]

```

data/external/scrapers/public_transport.py
```diff
@@ -0,0 +1,66 @@
...
+            station["distance"]=distance_to_building
+            results.append(station)
+    return results
+
+
+@cached_json("public_transport.json")
+def scrape_stations():
+    with Path("scrapers/MVV_HSTReport2212.csv").open("r") as file:
+        lines = csv.reader(file, delimiter=";")  
+        next(lines) # skip first line as it contains row names
```
CommanderStorm, line 39: ´csv.DictReader´ is imo a better option, as it protects you from index-lookup mistakes

data/external/scrapers/public_transport.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def avg(x1, x2):
+    return (x1 + x2) / 2
+
+
+def nearby(building_coords:tuple, stations: list[dict]) -> list[dict]:
+    results=[]
+    for station in stations:
+        distance_to_building=round(distance((station.get("lat"), station.get("lon")), (building_coords)).km,2)
```
CommanderStorm, line 26: Rounding here might be excessively strict and deprive the frontend from differentiating between close stations
=> Is there is a reasoning why rounding is better?

CommanderStorm (changes requested):
I have left some general advice and some structural advice.
In general, I am happy with this approach, but:

- I would like the building-matching code to live in, e.g. `processor/public_transport.py`
- Since you forked, there has been the development that the `scraper` ⇋ `processor` interface is now strongly typed via `pydantic.dataclass` 'es
  - I would like to see an interface like [`data/external/models/nat.py`](https://github.com/TUM-Dev/NavigaTUM/blob/main/data/external/models/nat.py) and a test for this interface like those in [`data/external/models/test_models.py`](https://github.com/TUM-Dev/NavigaTUM/blob/main/data/external/models/test_models.py`) 
- about downloading the csv file:
  - their documentation somewhat implies that this only includes bus stations. I don't know a station which is not a bus station too.  
  - if you look at the network tab of https://www.mvv-muenchen.de/fahrplanauskunft/fuer-entwickler/homepage-services/index.html, you can see the results here:
image
  (note that I find mixing base64 and urlencoding quite wild, but hey..)
=> I think there is more data we need to scrape given what [the API requires](https://www.mvv-muenchen.de/?eID=departuresFinder&action=get_departures&stop_id=de%3A09162%3A2&requested_timestamp=1687283704&lines=JmxpbmU9ZGRiJTNBOTJNMDElM0ElMjAlM0FIJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDElM0ElMjAlM0FSJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDIlM0ElMjAlM0FIJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDIlM0ElMjAlM0FSJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDMlM0ElMjAlM0FIJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDMlM0ElMjAlM0FSJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDQlM0ElMjAlM0FIJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDQlM0ElMjAlM0FSJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDYlM0ElMjAlM0FIJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDYlM0ElMjAlM0FSJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDclM0ElMjAlM0FIJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDclM0ElMjAlM0FSJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDglM0ElMjAlM0FIJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDglM0ElMjAlM0FSJTNBajIzJmxpbmU9c3dtJTNBMDEwMFUlM0FHJTNBSCUzQTAxMyZsaW5lPXN3bSUzQTAxMDBVJTNBRyUzQVIlM0EwMTMmbGluZT1zd20lM0EwMzEzMiUzQUclM0FIJTNBMDEz).

data/external/scrapers/public_transport.py
quarz12: The main benefit here is for better readability. I thought it would be best to see the distance in km with 2 decimal places. 
I am not sure what you mean with differentiating issues. each station is stored like this: 
```
      { "distance": 0.62,
        "global-id": "",
        "id": "8007",
        "lat": "48.1472252045499",
        "lon": "11.5605495171959",
        "name": "Brienner Str.",
        "ort": "M\u00fcnchen" 
      }
```
While rounding would make it impossible to sort by distance with 100% accuracy, I think there is enough information to differentiate stations.

quarz12:
> if you look at the network tab of https://www.mvv-muenchen.de/fahrplanauskunft/fuer-entwickler/homepage-services/index.html, you can see the results here:
> image
> (note that I find mixing base64 and urlencoding quite wild, but hey..)
> => I think there is more data we need to scrape given what [the API requires](https://www.mvv-muenchen.de/?eID=departuresFinder&action=get_departures&stop_id=de%3A09162%3A2&requested_timestamp=1687283704&lines=JmxpbmU9ZGRiJTNBOTJNMDElM0ElMjAlM0FIJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDElM0ElMjAlM0FSJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDIlM0ElMjAlM0FIJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDIlM0ElMjAlM0FSJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDMlM0ElMjAlM0FIJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDMlM0ElMjAlM0FSJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDQlM0ElMjAlM0FIJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDQlM0ElMjAlM0FSJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDYlM0ElMjAlM0FIJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDYlM0ElMjAlM0FSJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDclM0ElMjAlM0FIJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDclM0ElMjAlM0FSJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDglM0ElMjAlM0FIJTNBajIzJmxpbmU9ZGRiJTNBOTJNMDglM0ElMjAlM0FSJTNBajIzJmxpbmU9c3dtJTNBMDEwMFUlM0FHJTNBSCUzQTAxMyZsaW5lPXN3bSUzQTAxMDBVJTNBRyUzQVIlM0EwMTMmbGluZT1zd20lM0EwMzEzMiUzQUclM0FIJTNBMDEz).

This is about getting the timetables of departures, right? I thought it would be a good idea to first focus on just listing the stations. 
After doing some digging, I found out that the only information we really need is the stations id, everything else can be omitted. However the id has to have this format : ```de:09162:10``` so I will probably have to change to the other csv file.

data/external/scrapers/public_transport.py
quarz12: your proposed change would just change the data format from csv to json though? looking through all the stations and filtering them by distance does take some time, so I think it would be quite beneficial to have the results cached on the server. I do understand your concern about decoupling, however I feel like, in this scenario it is almost mandatory.

CommanderStorm: > looking through all the stations and filtering them by distance does take some time

That is fine. The main pipeline takes about 2m to complete. Most of this time is spend in image resizing.

Why I am not really concerned with the extra 78.4s is, that I don't think that the current algorithm (looping over every coordinate for every entry is n*m, which is pretty bad)

I think the standard techniques can help us a lot and should be able to get this into the regime where it is neglable:
- Your current model of the world is not necessary: using geodesic is accurate, but for small distances, assuming the earth is flat is fine.
  For example, the following runs in acceptable time.
  ```py
  def nearby(building_coords:tuple, stations: list[dict]) -> list[dict]:
    results=[]
    for station in stations:
        station_coords=(station.get("lat"), station.get("lon"))
        distance_to_building=great_circle(station_coords, building_coords).km
        if distance_to_building <= MAXDISTANCE:
            station["lat"]=str(station.get("lat"))  #decimal to string to allow json serialization
            station["lon"]=str(station.get("lon"))
            station["distance"]=geodesic(station_coords, building_coords).m
            results.append(station)
    return results
    ```
- bucketing/sorting: We really don't need to compare every item against every other station, finding a range via binary search should be enough.
- Caching: in development images are only resized when nessesary, this could be a factor here to, but I doubt that it is nessesary.

>  I do understand your concern about decoupling, however I feel like, in this scenario it is almost mandatory.

Another argument why separating them is a good idea is that we at this point (before running the pipeline) don't have a good overview of what buildings exist.

quarz12: I'm sorry, but I just dont understand at what point you would want to run this, then. If your suggestion is in python, then it must be part of the pipeline, as that is the only part written in python. Or am I misunderstanding something here?

CommanderStorm: Yes, the code snippet is a proposed alternative for the `nearby` function from your code.

quarz12: I got that, but what confuses me is that this proposed change:
```
@cached_json("public_transport.json")
def scrape_stations():
    with open("scrapers/MVV_HSTReport2212.csv", "r", newline='') as file:
        reader = csv.DictReader(file, delimiter=";")
        return [{
                    "id": int(line["﻿HstNummer"]),
                    "name":line["Name ohne Ort"],
                    "location":line["Ort"],
                    "global_id":line["Globale ID"],
                    "lat":Decimal(line["WGS84 X"].replace(",", ".")),
                    "lon":Decimal(line["WGS84 Y"].replace(",", ".")),
                } for line in reader if line["﻿HstNummer"]]
```
does not use that function at all. Should I make another json file that links building to station?

CommanderStorm: No, I don't think the matching needs to be statically cached.
I would like the building-matching code to live in `processor/public_transport.py`, as this has less coupling.

quarz12: I have now changed to using the other csv as only that one includes the stationID required for the timeplan js-lib. 
This one includes substations as well, so I put them in a list with in the parent station. I chose to return it as a dict for now, so it is easier to look up substations. let me know if you prefer a list though. 
Also, I added typing and a test.

data/external/scrapers/public_transport.py
CommanderStorm: Thinking about this a bit more:
Storing `0.xx km` here does not super make sense imo, I would prefer to display `xxx meters away` in the frontend, as that is probably more user-friendly than listing a bunch of `0.xxx km` ^^

quarz12: nearby now returns meters

quarz12:
I think the data collection part is ready for another review now. ```nearby``` is now a lot faster, but I am unsure of how to integrate it into the pipeline.

CommanderStorm (changes requested):
I have left some inline changes.
- Thinking about it, I think that binary search is overkill, as the code looks pretty unmaintanable.
  Suggesting this was probably a mistake.
  Sorry for this.
- In general, could you run the formatting suite `pre-commit run -a`
  (don't worry about extra mypy/... issues. I have just made the formatter/typechecker a lot stricter, which I still have to work through)

data/external/scrapers/public_transport.py
```diff
@@ -0,0 +1,47 @@
...
+WGS84X = "stop_lat"  # lat
+WSG84Y = "stop_lon"  # lon
+PARENT="parent_station"
+
+def avg(x1, x2):
+    return (x1 + x2) / 2
+
+@cached_json("public_transport.json")
+def scrape_stations():
+    with Path("scrapers/stops.txt").open("r") as file: #the file is from the zip file "Soll-Fahrplandaten" from https://www.mvv-muenchen.de/fahrplanauskunft/fuer-entwickler/opendata/index.html
```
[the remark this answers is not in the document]
quarz12: The file gets updated only once per year. Since a lot can change on the page in that time, I chose to not automate that. If you still think it is worth to do it automatically though, I can, of course, do that.

webclient/src/components/DetailsBuildingOverviewSection.vue
```diff
@@ -1,12 +1,60 @@
...
-import { useToggle } from "@vueuse/core";
 import type { components } from "@/api_types";
+import { useToggle } from "@vueuse/core";
 type BuildingsOverview = components["schemas"]["BuildingsOverview"];
 
 const props = defineProps<{
   readonly buildings?: BuildingsOverview;
 }>();
 
+import("https://www.mvv-muenchen.de/typo3conf/ext/sn_mvv_efa/Resources/Public/mvv-monitor/mvv-monitor.min.js");
```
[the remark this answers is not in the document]
quarz12: Tbh, I don't know. I tried putting it into another <script>, but that gives an error: A module cannot have multiple default exports.ts(2528). My knowledge of Vue is very limited, and I don't know of a different way to do it.

data/external/scrapers/public_transport.py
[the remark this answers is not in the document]
CommanderStorm: I think it is worth it.
Rationale: Long-term maintainability

data/processors/public_transport.py
```diff
@@ -0,0 +1,50 @@
...
+    if distance>0:
+       upper=current
+    else:
+      lower=current
+    current=lower+math.ceil((upper-lower)/2)
+  interval= _get_lat_interval(position,lst,current)
+  return lst[interval[0]:interval[1]+1]
+
+  
+def _get_lat_interval(position:tuple, lst:list[dict], point_in_interval:int)->tuple[int,int]:
```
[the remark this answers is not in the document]
quarz12: I suppose I could have better documented this:
assuming we search for stations nearby the x: 
image
```lat_bin_serarch``` first finds **a** station that is within a latitude distance of 1km (let that be the blue x):
image
that station then gets passed to ```get_lat_interval```, which goes left and right from that point, until the next point is not in  the 1km distance anymore. This interval is then returned: 
image
In my test, the binary approach is about 5 times faster, I guess which one is better depends on how many buildings this has to run for.

File: data/external/scrapers/public_transport.py [hunk 7]
```diff
@@ -0,0 +1,47 @@
...
+        lines = csv.DictReader(file, delimiter=",")  
+        stations={}
+        repeat_later=[] #when parent station is not already in dict
+        for line in lines:
+            if line[TYPE]:
+                stations.setdefault(line[STATIONID],{
+                    "id":line[STATIONID],
+                    "name":line[NAME],
+                    "lat":line[WGS84X].replace(",", "."),
+                    "lon":line[WSG84Y].replace(",", "."),
```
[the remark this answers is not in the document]
quarz12: about storing coordinates as float.
the reason I kept this as strings is this:
\>>> Decimal(1.23)
Decimal('1.229999999999999982236431605997495353221893310546875')
\>>> Decimal("1.23")
Decimal('1.23')
we could, however, say that we don't care about this slight inaccuracy and not use Decimals

data/processors/public_transport.py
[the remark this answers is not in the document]
quarz12: I just tried running the pipeline with your suggestion, and, assuming the way I implemented is correct, it finishes in ~1 sec. So I will go with that now.

PR #693, opened 2023-07-02
[MVV-1] Add close stations to the data model

CommanderStorm:
First Pull request in the #86 PR-Train
@quarz12 I have decided that this part of the code could be better maintained upstream
=> Rebasing of https://github.com/TUM-Dev/NavigaTUM/pull/663 will be nessesary

- Added a scraper to collect mvg stations
- add them to each room/building

## How to test this PR

1. Select a room
2. See if the stations are asigned correctly

## How has this been tested?

- see above

## Checklist

- [x] I have run the linter


quarz12:
I see, so I should open another pr for future changes? Also, I saw that you decided to compute nearby stations for every room. Are you sure that's a good idea? It seems like a lot of unnecessary redundancy to me.

CommanderStorm:
> > > Rebasing of #663 will be nessesary
> 
> I see, so I should open another pr for future changes?
> Edit: Oh, or did you mean that I should overwrite the branch with the new main

You can do as you like:
- Close #663 and opening a new, fresh PR (advantage: no rebasing+no huge discussion history)
- Rebase #663 onto master using `git rebase` and `git push --force-with-lease` (this is likely what you meant by overriding)
 
> Also, I saw that you decided to compute nearby stations for every room. Are you sure that's a good idea? It seems like a lot of unnecessary redundancy to me.

Yes, probably redundant, but I'd like Public transport to be available in every room (not only buildings)
=> Buildings are massive, and it might be significant (think main campus or MW) which bus station is displayed.
=> I experimented with it and fond that +8s is acceptable
=> deploying new data will now take 2m build + 10s deployment + ~30s rebuilding the Meilisearch index. This is not an issue because we are utilising Kubernetes rolling deployments. 
=> If something on the server is changed, the +8s are free, as compiling rust takes literal centuries.

PR #694, opened 2023-07-03
[MVV-3] Show close MVV stations

quarz12:
Second Pull request in the https://github.com/TUM-Dev/NavigaTUM/issues/86 PR-Train

- added train stations from another csv
- added MVV component to display nearby stations and their departures
image

## How to test this PR

1. Launch Subsystem XY
2. Click at
3. See changed XZ

## How has this been tested?

-
-

## Checklist

- [ ] I have updated the documentation / No need to update the documentation
- [x] I have run the linter


CommanderStorm:
(sorry, I accidentally rebased your branch too instead of merging into it)

CommanderStorm:
> Am I right when I say that type mvg is a tuple of Stations? I think a list would be more practical.

`mvg` is of type `Station[]` => list, but 
- the list includes at least one instance
- is readonly

Javascript [does not have tuples](https://ntgard.medium.com/tuples-in-javascript-cd33321e5277)

quarz12:
Ah, I see. That makes more sense. 

> Javascript [does not have tuples](https://ntgard.medium.com/tuples-in-javascript-cd33321e5277)

typescript does though ^^

CommanderStorm:
I apologise for responding slowly, the exam phase is taking its toll on me.

# Splitting this up into multiple PRs

In general, I would like this to be split up into at least two PRs (one modifying the `data` pipeline and one modifying the `webclient`).
Keeping changes atomic makes reviewing PRs much easier.

> Also, I realized that at some point some random files, that pre-commit changed, made it into a push sweat_smile. I hope that's not too big of an issue.

Actually, in the current state I cannot review this as there are many unrelated changes in here.
I would recommend the following:
- squash the changes you want into two Commits (one for the changes to the `data` pipeline and one for the `webclient`),
- drop any unrelated commits
- rebase onto `TUM-Dev/NavigaTUM:main` to resolve these changes properly
- force push to this and the open another branch whith the changes of `data`
- and change this PR (for `webclient`) to be based upon the second PR (for `data`)

> I do have a version that should work with live data, but for now the data is hard coded, as I don't have any other way of testing this feature. As mentioned before, I don't know rust, so I can't do anything about the server.

The server should not require a change.
The data of which stations are near to a room is already included in `poi?.mvg` mentioned above.
If you need any customisation, please tell me.

#  Design

Currently, the design is quite different from the design language we have in the rest of the repo.
I am not a fan of the current approach for multiple reasons and think this requires a rework.

The problems I have:

- **The sidebar**
I don't believe the current UI would be intuitive to new users:
image
- **Widget duplication**
Adding this widget means duplicating the widget above
=> I think this looks weird
- **design is not close**:
  This might sound pedantic to you (it is), but TUM is really concerned with its image.
  => Anything which is branded with TUM has to look like TUM
  => we have got into hot water with them recently and would like to keep the https://nav.tum.de domain 😉
  => Prominently utilising the MVV design is not an option
- us contacting the MVG API needs to be added to the be data protection policy (that is something I can do, if you don't feel comfortable with this) 

I would like this solution:
- I would like a simpler solution:
  => just a widget showing departures for a selection of Stations:   (this is a basic mock-up in paint, not a design mock-up 😉)
  image
- Design+functionality should be based on the `Rooms` search panel but with
  - variable icons and
  - more columns to accommodate the other items 
  - possibly no virtual scrolling (depending on how many departures there can be for central rooms)
- This should be split up
  - one PR adds the widget which accepts a list of stations as props and displays the departures
    - said PR handles contacting the MVG API (extracted from the API calls the MVG code makes) as well 
  - one PR to add filtering of stations (distance to the room is displayed here)

CC: @octycs (designer of Navigatum)

quarz12:
I finally found a way to get rid of all the unwanted changes. 🎉
About the design:
- in one of the later commits, I added units to the distance, maybe that makes it easier to understand the UI?
image

-  Getting the live data is quite easy like this: https://dev-redaktion.mvv-muenchen.de/?eID=departuresFinder&action=get_departures&stop_id=de%3A09162%3A1206&requested_timestamp=1689587573&lines=
However, I am everything but experienced in designing html and css, so I will probably only be able to create something practical. 
The same applies to the station buttons. I tried doing my own button list earlier, but it ended up looking way worse, so I decided to use what was already there.

octycs:
> * image

I'm not sure, what do you mean with "stations not filtered" here?

CommanderStorm:
> what do you mean with "stations not filtered" here?

Basically, just a dropdown button with you can specify the stations you want to filter for.

quarz12:
rebased to remove commits for the data pipeline as they were moved to #726

CommanderStorm:
@quarz12 I am sorry to have to tell you that the current implementation of this feature is incompatible with the design-language of the rest of the project.
Given that I don't really see this PR going anywhere in the current form, I am going to close it.

PR #716, opened 2023-07-09
Coordinate sanity check

### LINKED ISSUE #131: [Feature] Coordinate sainity check
as noticed in #129, some oviously wrong coordinates  (for example https://nav.tum.sexy/room/0104.U1.421) exist.
Ideally, we should notice (and then fix) these coordinates before we see them in production.

I would like to have a coordinate check. This could be implemented via a coverage approach `(max(lat)-min(lat))*(max(lon)-min(lon))` or via an absolute distance approach. 
The distance thresholds could be hardcoded or infered from the 95'th percentile of buildings in the same category

Not implementing such a check and waiting for user feedback, that indeed their shower is not located in africa..

**Additional context**
These cases seem to be really rare. 
There are most likely only a single digit amount of rooms, that are affected.

quarz12:
Resolves #131

- <strike>checks if coordinates are within this window. If not, throw exception (maybe deleting the coords attibute works too? I noticed the interactive map shows munich if theres no coords field provided)</strike>
image

- computes for each room the distance to its parent building and checks if it is <100m. If not, writes a warning.
- allows for overriding of building coordinates in coords_override.json
## How to test this PR

- run compile.py

## How has this been tested?

^

## Checklist

- [x] I have run the linter


CommanderStorm (changes requested):
- I think there are different approaches which could catch more errors than just overlaying a rectangle globally.
This check will, for example, not find if there is one room far, far outside the bounds of a building.
  A distance-based approach with either hardcoded or inferred distance from the room-corpus is a better approach:
  ```py
  if (distance_to_parent(data) > infered_maximum_distance_between_room_and_parent[...]):
  ```
  I would prefer an approach which infers the distance based on a percentile of the data * constant factor
- with the current approach, this won't build for 
  -> I would prefer these issues to be addressed in this PR.
  This can be done by
  - deleting the coordinates (override with no coordinate),
  - overriding with the parent,
  - overriding with a manually inferred/accurate coordinate or
  - making this a warning instead of a `Exception`

CommanderStorm:
(sorry, I accidentally rebased your branch too instead of merging into it)

quarz12:
I agree that this approach is quite inaccurate. I was not sure if I could assume building locations to be correct. That opens up way better solutions.

quarz12:
>I would prefer an approach which infers the distance based on a percentile of the data * constant factor

I am not sure I understand this right. Does this mean values should be accepted, if a certain percentile of rooms is close enough nearby?

CommanderStorm:
> I was not sure if I could assume building locations to be correct

All building locations can be assumed to be correct and are to some part manually sourced.
If there is a significant variation between the rooms and the building, that would still be interesting though ^^

quarz12:
> If there is a significant variation between the rooms and the building, that would still be interesting though ^^
well, there are a lot of rooms in 0510 that are about 2km away from their building. looking at the 0510 coords, I'd assume that that one is incorrect though. What do you think is a good maximum distance? I did some test runs, and it looks like all correct values are easily within 100m, so I'd suggest that.

quarz12:
I think this is now at a point where I can say that I am happy with the solution.
I know you gave me merge permissions on this repo, but I think I am still too new to this project to make this decision alone.

CommanderStorm (changes requested):
I have made some style-inline comments.
There are still quite a lot of linting issues in this PR. Please run the linter (included in `pre-commit`, see the Readme for further details) next time before requesting a review ^^

CommanderStorm:
> I know you gave me merge permissions on this repo, but I think I am still too new to this project to make this decision alone.  

Honestly, this was not intentional. I have changed a setting and this should not be possible anymore.
I was trying to give you the permission to run github-actions (like the CI) by yourself.
While for this PR this will not make a difference (https://github.com/TUM-Dev/NavigaTUM/pull/712 is not yet mergable), I thought the feedback if the docker files build could be valable feedback.

quarz12:
>Please run the linter (included in pre-commit

I thought, if I have installed that, it would automatically run? Is there a way to only run it for a single file? --all-files bloats my changes log with other files that then prevent me from switching branches until I delete them again.
>>I know you gave me merge permissions on this repo, but I think I am still too new to this project to make this decision alone.

>Honestly, this was not intentional. I have changed a setting and this should not be possible anymore.

I did think that was a bit much😅

data/processors/coords.py
```diff
@@ -118,9 +125,25 @@ def check_coords(input_data):
...
+def validate_coords(input_data):
+    for iid, data in input_data.items():
+        if not data["type"] =="room":
+            continue
+        coords=data["coords"]
+        parent_coords=input_data[data["parents"][-1]]["coords"]
+
+        if coords["lat"]==parent_coords["lat"] and coords["lon"]==parent_coords["lon"]:
+            continue
+
```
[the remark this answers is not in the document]
quarz12: for some buildings, when room and building coords are the same, ```sin(lat1) * sin(lat2) + cos(lat1) * cos(lat2) * cos(lon1 - lon2)``` returns a value like 1.0000000000000002. 
Which means acos raises an exception as it is bigger than 1.
I guess a better solution would be to use Decimals in the distance func.

data/processors/coords.py
```diff
@@ -2,6 +2,9 @@
 import logging
 
 import utm
+from utils import distance_via_great_circle
+
+MAX_DISTANCE_METERS_FROM_PARENT = 100
```
CommanderStorm, line 7: Pushing this to `200` resolves two false positives

```suggestion
MAX_DISTANCE_METERS_FROM_PARENT = 200
```

data/processors/coords.py
```diff
@@ -119,8 +122,31 @@ def check_coords(input_data):
             )
 
 
+def validate_coords(input_data):
+    for iid, data in input_data.items():
+        if not data["type"] == "room":
```
CommanderStorm, line 127:
```suggestion
        if data["type"] != "room":
```

data/processors/coords.py
```diff
@@ -119,8 +122,31 @@ def check_coords(input_data):
...
+            coords["lon"],
+            parent_coords["lat"],
+            parent_coords["lon"],
+        )
+
+        if distance_to_parent > MAX_DISTANCE_METERS_FROM_PARENT:
+            logging.warn(
+                f"{iid} {coords} is {distance_to_parent}m away from its parent "
+                + f"{parent_id} {parent_coords}. Please recheck if the coordinate makes sense",
+            )
```
CommanderStorm, line 144:
```suggestion
            logging.warning(
                f"{iid} {coords} is {distance_to_parent}m away from its parent {parent_id} {parent_coords}. "
                "Please recheck if the coordinate makes sense",
            )
```

CommanderStorm (approved):

data/processors/coords.py
```diff
@@ -119,8 +122,31 @@ def check_coords(input_data):
             )
 
 
+def validate_coords(input_data):
```
CommanderStorm, line 125: Added a method docstring
```suggestion
def validate_coords(input_data):
    """Check that coordinates are not too far away from their parent"""
```

data/processors/coords.py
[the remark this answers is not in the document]
CommanderStorm: I like the current solution much better 👍🏻

CommanderStorm:
Thank you for implementing this ❤️

PR #726, opened 2023-07-20
[MVV-2] Attach train station information to the public transport information

quarz12:
Fixes #

- data part of #694
- adds data from second csv, that contains train stations, too. previous pr was only bus stations

## How to test this PR

1. models/testmodels.py

## How has this been tested?

- running main.py

## Checklist

- [x] I have run the linter

### DIFF
--- a/data/external/scrapers/public_transport.py
+++ b/data/external/scrapers/public_transport.py
@@ -18,7 +18,7 @@ def _load_bus_stations() -> dict:
                 stations.setdefault(
                     line["stop_id"],
                     {
-                        "id": line["stop_id"],
+                        "station_id": line["stop_id"],
                         "name": line["stop_name"],
                         "lat": float(line["stop_lat"]),
                         "lon": float(line["stop_lon"]),
@@ -27,7 +27,7 @@ def _load_bus_stations() -> dict:
                 )
             else:
                 sub_station = {
-                    "id": line["stop_id"],
+                    "station_id": line["stop_id"],
                     "name": line["stop_name"],
                     "lat": float(line["stop_lat"]),
                     "lon": float(line["stop_lon"]),
@@ -57,7 +57,7 @@ def _load_train_stations(stations: dict) -> list:
                 stations.setdefault(
                     line["Globale ID"],
                     {
-                        "id": line["Globale ID"],
+                        "station_id": line["Globale ID"],
                         "name": line["Name ohne Ort"],
                         "lat": float(line["WGS84 X"].replace(",", ".")),
                         "lon": float(line["WGS84 Y"].replace(",", ".")),
@@ -67,7 +67,7 @@ def _load_train_stations(stations: dict) -> list:
             else:
                 parent_id = ".".join(line["Globale ID"].split(":")[0:3])
                 sub_station = {
-                    "id": line["Globale ID"],
+                    "station_id": line["Globale ID"],
                     "name": line["Name ohne Ort"],
                     "lat": float(line["WGS84 X"].replace(",", ".")),
                     "lon": float(line["WGS84 Y"].replace(",", ".")),
--- a/data/external/scrapers/public_transport.py
+++ b/data/external/scrapers/public_transport.py
@@ -38,7 +38,7 @@ def _load_bus_stations(stations: dict) -> None:
                 "parent": line["parent_station"],
             }
             if not sub_station["parent"]:
-                sub_station["parent"]=":".join(line["stop_id"].split(":")[:3])
+                sub_station["parent"] = ":".join(line["stop_id"].split(":")[:3])
 
             if parent := stations.get(line["parent_station"]):
                 parent["sub_stations"].append(sub_station)
@@ -50,7 +50,8 @@ def _load_bus_stations(stations: dict) -> None:
             parent["sub_stations"].append(sub)
         else:
             if sub["station_id"]:
-                logging.warn(f"{sub['name']} with id {sub['station_id']} has no parent in our data")
+                logging.warning(f"{sub['name']} with id {sub['station_id']} has no parent in our data")
+
 
 def _load_train_stations(stations: dict) -> None:
     """Load the bus stations from the MVV_HST_REPORT data and add them to stations dict"""
--- a/data/external/scrapers/public_transport.py
+++ b/data/external/scrapers/public_transport.py
@@ -2,99 +2,96 @@ import csv
 from zipfile import ZipFile
 
 from external.scraping_utils import _download_file, CACHE_PATH, cached_json
-
-MVV_GTFS_URL = "https://www.mvv-muenchen.de/fileadmin/mediapool/02-Fahrplanauskunft/03-Downloads/openData/mvv_gtfs.zip"
-MVV_HST_REPORT_URL = "https://www.mvv-muenchen.de/fileadmin/mediapool/02-Fahrplanauskunft/03-Downloads/openData/MVV_HSTReport2212.csv"  # train/tram stations + some bus stations
+MVV_OPENDATA_URL = "https://www.mvv-muenchen.de/fileadmin/mediapool/02-Fahrplanauskunft/03-Downloads/openData"
+MVV_GTFS_URL = f"{MVV_OPENDATA_URL}/mvv_gtfs.zip"
+MVV_HST_REPORT_URL = f"{MVV_OPENDATA_URL}/MVV_HSTReport2212.csv"  # train/tram stations + some bus stations
 PUBLIC_TRANSPORT_CACHE_PATH = CACHE_PATH / "public_transport"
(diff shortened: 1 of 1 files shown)


data/external/scrapers/public_transport.py
```diff
@@ -47,8 +42,59 @@ def scrape_stations():
...
+        return stations.values()
+
+
+@cached_json("public_transport.json")
+def scrape_stations():
+    """Scrape the stations from the MVV GTFS data and return them as a list of dicts"""
+    _download_file(MVV_GTFS_URL, PUBLIC_TRANSPORT_CACHE_PATH / "fahrplandaten.zip")
+    with ZipFile(PUBLIC_TRANSPORT_CACHE_PATH / "fahrplandaten.zip") as file_zip:
+        file_zip.extract("stops.txt", PUBLIC_TRANSPORT_CACHE_PATH)
+    _download_file(MVV_HST_REPORT_URL, PUBLIC_TRANSPORT_CACHE_PATH / "train_stations.csv")
```
CommanderStorm, line 93: I think these files should be downloaded in the functions where they are used given that they are only used there.

CommanderStorm (changes requested):

data/external/scrapers/public_transport.py
```diff
@@ -4,16 +4,11 @@
 from external.scraping_utils import _download_file, CACHE_PATH, cached_json
 
 MVV_GTFS_URL = "https://www.mvv-muenchen.de/fileadmin/mediapool/02-Fahrplanauskunft/03-Downloads/openData/mvv_gtfs.zip"
+MVV_HST_REPORT_URL = "https://www.mvv-muenchen.de/fileadmin/mediapool/02-Fahrplanauskunft/03-Downloads/openData/MVV_HSTReport2212.csv"  # train/tram stations + some bus stations
```
CommanderStorm, line 7: Above breaks the linting due to the lenght of the line
```suggestion
MVV_OPENDATA_URL = "https://www.mvv-muenchen.de/fileadmin/mediapool/02-Fahrplanauskunft/03-Downloads/openData"
MVV_GTFS_URL = f"{MVV_OPENDATA_URL}/mvv_gtfs.zip"
MVV_HST_REPORT_URL = f"{MVV_OPENDATA_URL}/MVV_HSTReport2212.csv"
```

CommanderStorm, line 7: This is named `MVV_HSTReport2212.csv` => Will we need to update this in the future?

data/external/scrapers/public_transport.py
```diff
@@ -47,8 +42,59 @@ def scrape_stations():
...
+                    {
+                        "station_id": line["Globale ID"],
+                        "name": line["Name ohne Ort"],
+                        "lat": float(line["WGS84 X"].replace(",", ".")),
+                        "lon": float(line["WGS84 Y"].replace(",", ".")),
+                        "sub_stations": [],
+                    },
+                )
+            else:
+                parent_id = ".".join(line["Globale ID"].split(":")[0:3])
```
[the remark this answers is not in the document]
quarz12: no idea how that got there, ```:``` is correct. I even had it in earlier versions.

data/external/scrapers/public_transport.py
quarz12: probably, yes

data/external/scrapers/public_transport.py
```diff
@@ -47,8 +42,59 @@ def scrape_stations():
...
+
+@cached_json("public_transport.json")
+def scrape_stations():
+    """Scrape the stations from the MVV GTFS data and return them as a list of dicts"""
+    _download_file(MVV_GTFS_URL, PUBLIC_TRANSPORT_CACHE_PATH / "fahrplandaten.zip")
+    with ZipFile(PUBLIC_TRANSPORT_CACHE_PATH / "fahrplandaten.zip") as file_zip:
+        file_zip.extract("stops.txt", PUBLIC_TRANSPORT_CACHE_PATH)
+    _download_file(MVV_HST_REPORT_URL, PUBLIC_TRANSPORT_CACHE_PATH / "train_stations.csv")
+    stations = _load_bus_stations()
+    stations = _load_train_stations(stations)
```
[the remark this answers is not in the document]
quarz12: That was a bit messy, got to admit that. 
I have changed both methods to add to a parameter dict. It now looks like this:
```
stations = {}
_load_bus_stations(stations)
_load_train_stations(stations)
```
The reason for keeping the data in a dict is, that stations in the train data can have parents in the bus data. 
Your suggestion of merging dicts might work, too, but this one definitely does.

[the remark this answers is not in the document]
CommanderStorm: Have you checked this assumption?
I would assume this to be the other way around (bus stations belonging to train stations being a thing)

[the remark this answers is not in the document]
quarz12: in general, parent stations are neither train nor bus stations, but the collection of their substations. If you request departures for a parent, you get all the childrens departures. If a parent only has 1 child, the child becomes the parent.

Realistically, I dont think there would be a child station without its parent in either csv. That just wouldn't make sense. 
If you want, i'll change it to dict merging. But this works just as well, so, from my perspective, there's not really a point in doing so.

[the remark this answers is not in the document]
CommanderStorm: > Realistically, I dont think there would be a child station without its parent in either csv.
> That just wouldn't make sense.

Having assumptions about foreign data is what kills you in data science.
=> If you have tested that this assumption holds true, we can merge this PR. (I am fine with calling+merging it as is, debugging such an issue later will be a nightmare though => let's do so now)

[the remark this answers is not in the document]
quarz12: I added checks/warnings for this case. also found some bugs I fixed, but now all the stations are getting added to our data.

data/external/scrapers/public_transport.py
```diff
@@ -1,54 +1,105 @@
...
+                parent["sub_stations"].append(sub_station)
+            else:
+                repeat_later.append(sub_station)
+
+    for sub in repeat_later:
+        if parent := stations.get(sub["parent"]):
+            parent["sub_stations"].append(sub)
+        else:
+            if sub["station_id"]:
+                logging.warn(f"{sub['name']} with id {sub['station_id']} has no parent in our data")
```
CommanderStorm, line 53:
```suggestion
                logging.warning(f"{sub['name']} with id {sub['station_id']} has no parent in our data")
```

data/external/scrapers/public_transport.py
```diff
@@ -1,54 +1,105 @@
...
+        else:
+            sub_station = {
+                "station_id": line["stop_id"],
+                "name": line["stop_name"],
+                "lat": float(line["stop_lat"]),
+                "lon": float(line["stop_lon"]),
+                "parent": line["parent_station"],
+            }
+            if not sub_station["parent"]:
+                sub_station["parent"]=":".join(line["stop_id"].split(":")[:3])
```
CommanderStorm, line 41:
```suggestion
                sub_station["parent"] = ":".join(line["stop_id"].split(":")[:3])
```

data/external/scrapers/public_transport.py
```diff
@@ -1,54 +1,105 @@
...
+            else:
+                repeat_later.append(sub_station)
+
+    for sub in repeat_later:
+        if parent := stations.get(sub["parent"]):
+            parent["sub_stations"].append(sub)
+        else:
+            if sub["station_id"]:
+                logging.warn(f"{sub['name']} with id {sub['station_id']} has no parent in our data")
+
```
CommanderStorm, line 54:
```suggestion


```

data/external/scrapers/public_transport.py
```diff
@@ -1,54 +1,106 @@
...
+            if parent := stations.get(parent_id):
+                parent["sub_stations"].append(sub_station)
+            else:
+                repeat_later.append(sub_station)
+    for sub in repeat_later:
+        if parent := stations.get(sub["parent"]):
+            parent["sub_stations"].append(sub)
+        else:
+            if sub["station_id"]:
+                logging.warn(f"{sub['name']} with id {sub['station_id']} has no parent in our data")
```
CommanderStorm, line 93:
```suggestion
                logging.warning(f"{sub['name']} with id {sub['station_id']} has no parent in our data")
```

data/external/scrapers/public_transport.py
```diff
@@ -1,54 +1,106 @@
 import csv
+import logging
 from zipfile import ZipFile
 
 from external.scraping_utils import _download_file, CACHE_PATH, cached_json
```
CommanderStorm, line 5:
```suggestion
from external.scraping_utils import _download_file, CACHE_PATH, cached_json

```

PR #1154, opened 2024-05-06
[breaking, api] Removed `/api/get/root` route

CommanderStorm:

- Currently the api response for the `root` route is quite different compared to other routes
- => removing it allows for some refactoring nessesary for the calendar PR

@jakobkoerber
The TCA does not use this subfeature/weirdness, but still a minor headsup about this API removal ^^
I don't think this really matters to you.

## How to test this PR

1. See that the 

## How has this been tested?

- See above

## Checklist

- [x] I have run the linter


jakobkoerber:
@CommanderStorm is there any chance this affects the TCA? We're using these endpoints:

```
case NavigaTumApiServiceSearch _:
        return "api/search";
      case NavigaTumApiServiceDetails details:
        return "api/get/${details.id}";
      case NavigaTumApiServiceImages images:
        return "cdn/maps/roomfinder/${images.id}";
      case NavigaTumApiServiceOverlayImages overlayImages:
        return "cdn/maps/roomfinder/${overlayImages.id}";
    }
```

CommanderStorm:
Since the TCA does not show root (that is the front-page you currently see with "Standorte") and this is virtual root of all other entries => not visible in any other room, not linkable  due to different data and not visible in the search
=> The TCA is not affected.

PR #1296, opened 2024-07-11
Postgis as a postgres replacement

CommanderStorm:
Moves us closer to the OSM goal

- Postgres -> PostGIS (a distribution of postgres)

## How to test this PR

1. launch the system
2. see that it still works

## How has this been tested?

- Letting it launch, import OSM data

## Checklist

- [x] I have run the linter


.gitignore
```diff
@@ -1,3 +1,4 @@
 venv
 __pycache__
 .idea
+*.pbf
```
CommanderStorm, line 4:
```suggestion
*.pbf

```

PR #1656, opened 2024-11-07
Removed Weilheim from Versuchsanstalt Obernach

CommanderStorm:
This should have been a commit to master, so be it


CommanderStorm:
split by cases and merged into master

PR #1685, opened 2024-11-19
add a tool to convert CAD to OSM

CommanderStorm:
Currentlly only a draft


CommanderStorm:
this approach is not working => lets try a new one

PR #1766, opened 2025-01-15
added normalised roomcodes to search for

### LINKED ISSUE #1733: Searching for Substrings of the room name / number („Raumkennung“)
When searching for the room 1.221 in the Hauptgebäude, searching for 1.221 will return no useful results. You either need to search for 1221 (Architektenname) or 01.221.

When searching for a room, consider all superstrings of that room in the Architektenname and the Raumkennung

I have not considered any alternatives.

**Additional context**
Current behavior: 

|  |  |
|--------|--------|
| image | image |
| image | image |

BaumBanane:
Fixes #1733 

## Proposed Change(s)

- added room_code, room_code_normalised and arch_name_normalised to the exported room data
- added these attributes to the searchable attributes in meilisearch.rs, removed ms_id (now replaced with room_code containing '.' as separator instead of '-')
- added normalise_id to export.py to calculate the normalised room_code_normalised and arch_name_normalised
- fixed a tiny typo in the lexer

## Checklist

- Documentation
    - [x] I have updated the documentation
    - [ ] No need to update the documentation

### DIFF
--- a/data/processors/export.py
+++ b/data/processors/export.py
@@ -41,6 +41,18 @@ def unlocalise(value: str | list[Any] | dict[str, Any]) -> Any:
     raise ValueError(f"Unhandled type {type(value)}")
 
 
+def normalise_id(_id: str) -> str:
+    parts=_id.split(".")
+    for i in range(0, len(parts)):
+        parts[i]=parts[i] .lstrip('0')
+        if not parts[i]:
+            parts[i]="0"
+    result: str = "";
+    for part in parts:
+        result+=part;
+    return ".".join(parts)
+
+
 def export_for_search(data: dict) -> None:
     """Export a subset of the data for the /search api"""
     export = []
@@ -75,9 +87,12 @@ def export_for_search(data: dict) -> None:
                 # MeiliSearch requires an id without "."
                 # also this puts more emphasis on the order (because "." counts as more distance)
                 "ms_id": _id.replace(".", "-"),
+                "room_code": _id,
+                "room_code_normalised": normalise_id(_id),
                 "id": _id,  # not searchable
                 "name": entry["name"],
                 "arch_name": entry.get("arch_name"),
+                "arch_name_normalised": normalise_id(entry.get("arch_name")),
                 "type": entry["type"],
                 "type_common_name": entry["type_common_name"],
                 "facet": {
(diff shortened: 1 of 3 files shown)


CommanderStorm:
PSA:
Never open a PR from your protected branch (`main`/`master`).
This prohibits "Maintainers are allowed to edit this pull request".

(it might also makes CI not run, but that is a different issue which might be unrelated)

data/processors/export.py
```diff
@@ -40,6 +40,18 @@ def unlocalise(value: str | list[Any] | dict[str, Any]) -> Any:
...
+    parts=_id.split(".")
+    for i in range(0, len(parts)):
+        parts[i]=parts[i] .lstrip('0')
+        if not parts[i]:
+            parts[i]="0"
+    result: str = "";
+    for part in parts:
+        result+=part;
+    return ".".join(parts)
+
```
CommanderStorm, line 54: (not a nit by me, but rather how ruff wants this formatted)

```suggestion

def normalise_id(_id: str) -> str:
    """Remove leading zeros from all point-separated parts of input string"""
    parts = _id.split(".")
    for i in range(0, len(parts)):
        parts[i] = parts[i].lstrip("0")
        if not parts[i]:
            parts[i] = "0"
    result: str = ""
    for part in parts:
        result += part
    return ".".join(parts)

```

data/processors/export.py
```diff
@@ -41,6 +41,19 @@ def unlocalise(value: str | list[Any] | dict[str, Any]) -> Any:
...
+    """Remove leading zeros from all point-separated parts of input string"""
+    parts = _id.split(".")
+    for i in range(0, len(parts)):
+        parts[i] = parts[i].lstrip("0")
+        if not parts[i]:
+            parts[i] = "0"
+    result: str = ""
+    for part in parts:
+        result += part
+    return ".".join(parts)
```
CommanderStorm, line 54: This might be nice to know:
Iterating with indexes in python is really bad performance wise.
For cases like this, list comprehensions are a bit more readable and a lot more performant.
(not like performance matters here, but )

```suggestion
    parts = [part.lstrip("0") or "0" for part in _id.split(".")]
    return ".".join(parts)
```

CommanderStorm:
lets test it in prod ^^

PR #1931, opened 2025-04-16
Calendar darker on hover

StratosDelis:
Fixes # Not a specific issue
Now the "Calendar" Icon gets darker (and lighter accordingly while using the dark theme) when the user hovers over it, just like the "Share" and the "Report" Icons. It is not really significant but once I noticed it, it was always quite irritating.
In the README.md I just corrected some small lower/upper case inconsistencies and added some full stops wherever they should be at.
## Proposed Change(s)

- It is a very small fix, I do not think there is something else that should be changed for this.

## Checklist

- Documentation
    - [ ] I have updated the documentation
    - [x] No need to update the documentation

Before

image

After

image

### DIFF
--- a/webclient/README.md
+++ b/webclient/README.md
@@ -78,15 +78,15 @@ Our CSS framework is [Tailwind](https://tailwindcss.com/).
 
 ```plain
 webclient
-├── public/        # 🠔 Static assets such as icons, which cannot get inlined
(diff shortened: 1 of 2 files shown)


CommanderStorm (approved):
Thanks ❤️

PR #2006, opened 2025-06-08
Maintain language preference from building -> room

alkminisarigeorgiou:
## The Issue
If the default language (German) is switched, when navigating through the app (specifically during visiting a room page from its corresponding building page), the language automatically switches back to the default and it has to be manually switched again. (Ex. Image 1)

This proposed change fixes this issue so that the language preference is kept and doesn't need to be configured constantly.
(Ex. Image 2)

## Before
https://github.com/user-attachments/assets/61c58581-e7f6-4f0b-ba13-24007676fe89

## After
https://github.com/user-attachments/assets/e19fdbe4-a59d-48a0-befd-c5071a483dc7

## Checklist

- Documentation
    - [ ] I have updated the documentation
    - [x] No need to update the documentation

### DIFF
--- a/webclient/app/components/DetailsRoomOverviewSection.vue
+++ b/webclient/app/components/DetailsRoomOverviewSection.vue
@@ -17,6 +17,9 @@ const props = defineProps<{
   readonly rooms?: RoomsOverviewResponse | null;
 }>();
 
+// Keep language preference when going from building page to room page
+const localePath = useLocalePath();
+
 const { t } = useI18n({ useScope: "local" });
 const selectedUsage = ref(-1);
 const search = ref("");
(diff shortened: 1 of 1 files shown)


webclient/app/components/DetailsRoomOverviewSection.vue
```diff
@@ -17,6 +17,9 @@ const props = defineProps<{
   readonly rooms?: RoomsOverviewResponse | null;
 }>();
 
+// Keep language preference when going from building page to room page
+const localePath = useLocalePath();
+
```
CommanderStorm, line 22:
```suggestion
```

webclient/app/components/DetailsRoomOverviewSection.vue
```diff
@@ -171,7 +174,7 @@ const { list, containerProps, wrapperProps } = useVirtualList<RoomsOverviewUsage
             <NuxtLinkLocale
               v-for="(room, index) in list"
               :key="index"
-              :to="`/view/${room.data.id}`"
+              :to="localePath(`/view/${room.data.id}`)"
```
CommanderStorm, line 177:
```suggestion
              :to="`/view/${room.data.id}`"
```

CommanderStorm (approved):
I tested removing the `external` prop and this works. Thanks for looking into this, even though the exact fix is slightly different

alkminisarigeorgiou:
> I tested removing the `external` prop and this works. Thanks for looking into this, even though the exact fix is slightly different

Haha you beat me to race ! I also just tested your suggestion and it actually is the root cause. Thanks a lot for your quick feedback.

PR #2218, opened 2025-09-20
Motis frontend v1

CommanderStorm:
implements a motis frontend


CommanderStorm:
merging to have the api change in master

PR #2524, opened 2026-01-03
chore: better local dev

CommanderStorm:
this PR makes local development simpler by shifting us from the "everything goes via the web" to a "maybe consider local files, fallback to the web" achitecture.

It also comes with tests.


data/Dockerfile
```diff
@@ -30,14 +30,11 @@ RUN cp external/results/public_transport.parquet output/public_transport.parquet
     && cp -r output/maps/overlays output/maps/overlay
 # last renaming is to remain backwards compatible
 
-# compress data (only using gzip, because brotli on ngnix is a royal pain)
+# Compress data (only using gzip, because brotli is complex)
 RUN gzip --force --keep --recursive output/
```
CommanderStorm:
```suggestion
```

data/Dockerfile
```diff
@@ -30,14 +30,11 @@ RUN cp external/results/public_transport.parquet output/public_transport.parquet
     && cp -r output/maps/overlays output/maps/overlay
 # last renaming is to remain backwards compatible
```
CommanderStorm:
```suggestion
```

PR #2681, opened 2026-02-07
update snapshots

CommanderStorm:
Somehow this sneaked into prod, no clue


server/src/search_executor/snapshots/navigatum_server__search_executor__test__bad_queries-15.snap
```diff
@@ -1,22 +1,11 @@
 ---
-source: src/search_executor/mod.rs
+source: server/src/search_executor/mod.rs
```
CommanderStorm:
```suggestion
source: src/search_executor/mod.rs
```

PR #2899, opened 2026-04-26
fix(server): preserve underscores in search filter slugify

CommanderStorm:
leads to not letting the search UI work correctly

### DIFF
--- a/server/src/routes/search.rs
+++ b/server/src/routes/search.rs
@@ -730,18 +730,14 @@ mod tests {
     fn filter_preserves_underscores_in_type_values() {
         // Regression: previously `_` was rewritten to `-`, so `virtual_room` became
         // `virtual-room` and never matched any document in the index.
-        let filter = build_meilisearch_filter(
-            &[],
-            &[],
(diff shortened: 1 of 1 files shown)


server/src/routes/search.rs
```diff
@@ -725,6 +726,22 @@ mod tests {
...
+            build_meilisearch_filter(&[], &[], &["room".to_string(), "virtual_room".to_string()]);
+        assert_eq!(filter, r#"(type IN ["room", "virtual_room"])"#);
+    }
+
+    #[test]
+    fn filter_preserves_underscores_in_parent_values() {
+        let filter = build_meilisearch_filter(&[], &[], &["joined_building".to_string()]);
+        assert!(filter.contains("joined_building"));
+        assert!(!filter.contains("joined-building"));
+    }
```
CommanderStorm:
```suggestion
    #[test]
    fn filter_preserves_underscores_in_parent_values() {
        let filter = build_meilisearch_filter(&[], &[], &["joined_building".to_string()]);
        assert!(filter.contains("joined_building"));
    }
```

PR #2920, opened 2026-04-26
fix(server): load new upstream rows during bootstrap

CommanderStorm:
fix the server joining the wrong way


data/processors/areatree/process.py
```diff
@@ -1,5 +1,6 @@
 import logging
 import re
+from pathlib import Path
```
CommanderStorm:
```suggestion
```

data/processors/tumonline.py
```diff
@@ -210,6 +210,7 @@ def _building_names_equivalent(
 _BUILDING_TYPES = ["building", "joined_building"]
 
 
+
```
CommanderStorm:
```suggestion
```

PR #3062, opened 2026-06-04
feat(server,data): motis-sourced nearby public-transport endpoint

### LINKED ISSUE #86: [Feature] Show close MVV stations

CommanderStorm:
Closes #86 (backend half — frontend ships in a follow-up PR).

## Summary

Replaces the DELFI/GTFS station catalogue with a motis-sourced one and exposes the new shape at `/api/locations/{id}/nearby`. The frontend consumer that actually renders the nearby section lives in a sibling PR and depends on this one.

## Data pipeline

`data/external/scrapers/public_transport.py` no longer pulls the manually-downloaded zHV-Gesamt zip. Instead it:

1. Reads `data/sources/coordinates.csv` (~925 TUM coordinates).
2. Single-linkage clusters them with eps = 5 km via union-find on a polars cross-join, producing 6 padded bboxes (Munich-core, Garching, Weihenstephan, Ottobrunn, Straubing, Heilbronn).
3. Hits motis `/api/v1/map/stops` per bbox over a shared `requests.Session` with a 1.5 s courtesy throttle (agreed with motis maintainers) and 3× exponential retry, fail-hard.
4. Folds platform rows sharing `(parentId, name)` into one parent station (centroid coords, unioned modes); stops without `parentId` are kept as-is keyed by `stopId`.
5. Writes `public_transport.parquet` with columns `id, name, modes: list[enum], lat, lon`.

End-to-end: 1692 stations, modes distribution `bus 1359, tram 185, subway 58, ride_sharing 57, suburban 49, regional_rail 36, highspeed_rail 28, long_distance 19, odm 16, night_rail 10, coach 8`.

## Server

- New migration drops `transportation_stations` and recreates it as `(id TEXT PK, name TEXT NOT NULL, modes TEXT[] NOT NULL DEFAULT '{}', coordinate Point NOT NULL)` + GIST index. Parent join is gone; the scraper synthesises parents.
- `nearby_handler` response: `modes: Vec<ModeResponse>` (lowercase snake_case, matching the existing `TransitStopType` / `TravelMode` style). `parent_id` / `parent_name` are gone.
- `ModeResponse` (the existing motis route enum) is extended with `Suburban` and `RideSharing` — motis serves them on `/map/stops` but the progenitor-generated `Mode` doesn't list them — plus a `parse_or_other()` helper that folds unknown / future motis strings into `Other`. `mod motis` → `pub(crate) mod motis` so the enum can be shared.
- sqlx cache regenerated with `--tests` (also restores the previously-dropped `seed_de` cache entry).

## Frontend (only the contract pieces)

The actual `DetailsNearbyTransportSection.vue` and its mount land in a follow-up. This PR carries only:

- `MotisTransitModeIcon.vue` extended to handle `suburban` and `ride_sharing`.
- `webclient/app/api_types/index.ts` regenerated from `openapi.yaml` so the type union matches the new `ModeResponse`.

## Test

- New case in `tests/specs/server.api.spec.ts`: hits `/api/locations/mi/nearby` and asserts the response items have a non-empty `modes: snake_case[]` (members of `ModeResponse`), no `parent_id` / `parent_name`, distance in `(0, 1000)`, and are sorted by ascending distance.

## Test plan

- [x] `cargo check --workspace --tests` (offline, with regenerated sqlx cache)
- [x] `cargo sqlx prepare -- --tests` passes against a fresh postgis container
- [x] `uv run pytest external/` — 26 passed (drift-gate validates the regenerated parquet)
- [x] `uv run ruff check` clean
- [x] `pnpm run type-check` clean (regenerated `app/api_types/index.ts`)
- [x] Scraper end-to-end against live transitous — 6 bboxes, 1910 platform rows → 1692 stations
- [ ] Watch the initial production migration for the brief restart-loop window while old containers drain (accepted in the design)


CommanderStorm:
Heads up: scope split into two PRs per request — this one is backend-only, the frontend section lives in #3065 (draft until this lands).

CommanderStorm:
Force-pushed: the scraper's parent-folding logic was incomplete — motis only emits `parentId` on subway platforms, not bus tracks, so 5 separate rows survived for stations like "Garching, Forschungszentrum". Fixed by folding by `name` (already city-prefixed in German GTFS) instead of `(parentId, name)`. Parquet shrank from 1692 → 728 rows. The Playwright test now asserts the fold actually happened: it expects exactly one "Garching, Forschungszentrum" row in `/api/locations/mi/nearby` with both `subway` and `bus` in modes — that's the regression I would actually want this test to catch.

server/migrations/20260604120000_transport_modes.sql
```diff
@@ -0,0 +1,13 @@
+DROP TABLE IF EXISTS transportation_stations CASCADE;
+
+CREATE TABLE transportation_stations
+(
+    id         TEXT PRIMARY KEY,
+    name       TEXT   NOT NULL,
+    modes      TEXT[] NOT NULL DEFAULT '{}',
```
CommanderStorm:
```suggestion
    modes      TEXT[] NOT NULL CHECK (cardinality(modes) > 0),
```

PR #3065, opened 2026-06-04
feat(webclient): nearby public-transport section on detail pages

### LINKED ISSUE #86: [Feature] Show close MVV stations

CommanderStorm:
Closes #86


webclient/app/components/DetailsContentSidebar.vue
```diff
@@ -214,6 +214,9 @@ const suggestLocationFix = () => {
       <ClientOnly>
         <LazyDetailsRoomOverviewSection :rooms="data.sections?.rooms_overview"/>
       </ClientOnly>
+      <ClientOnly>
```
CommanderStorm:
```suggestion
```

PR #3458, opened 2026-07-15
docs(server,webclient): surface the blind-profile public-transit fallback

### LINKED ISSUE #3454: Surface the blind-profile fallback instead of silently routing as standard walking
## What to build

The routing API accepts a blind pedestrian type, but MOTIS does not currently support blind-specific routing, so the request is silently collapsed to the standard foot profile. Users who select the blind profile get standard walking routes without any indication that their choice had no effect.

Make this limitation explicit instead of silent: state in the OpenAPI documentation of the pedestrian type that the blind profile currently routes as standard walking because the routing engine lacks a blind profile, and inform users in the UI when they select it. The actual routing behaviour stays unchanged; this slice is about honesty, not new engine capability.

## Acceptance criteria

- [ ] The OpenAPI description of the pedestrian type states that the blind profile currently falls back to standard walking due to the engine limitation.
- [ ] A user who selects the blind profile in the UI sees a hint about the fallback.
- [ ] The hint copy exists in both German and English.
- [ ] Routing behaviour itself is unchanged.

## Blocked by

None - can start immediately.

CommanderStorm:
Closes #3454 — states in the OpenAPI schema description and a bilingual preferences-popup hint that the blind pedestrian profile falls back to standard walking on public transit routes (only the MOTIS path collapses it; walking routes via Valhalla honour the profile), leaving routing behaviour unchanged.

### DIFF
--- a/server/src/routes/maps/route/mod.rs
+++ b/server/src/routes/maps/route/mod.rs
@@ -228,11 +228,6 @@ struct RoutingRequest {
 }
 
 /// Does the user have specific walking needs?
-///
-/// **Note:** For public transit routes, `blind` currently falls back to standard walking
-/// because the routing engine ([MOTIS](https://github.com/motis-project/motis)) does not
-/// offer a blind-specific pedestrian profile yet.
-/// Walking routes honour the profile via more detailed route instructions.
 #[derive(Deserialize, Debug, Default, Clone, Copy, PartialEq, Eq, utoipa::ToSchema)]
 #[serde(rename_all = "snake_case")]
 enum PedestrianTypeRequest {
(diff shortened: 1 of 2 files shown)


server/src/routes/maps/route/mod.rs
```diff
@@ -228,6 +228,11 @@ struct RoutingRequest {
 }
 
 /// Does the user have specific walking needs?
+///
+/// **Note:** For public transit routes, `blind` currently falls back to standard walking
+/// because the routing engine ([MOTIS](https://github.com/motis-project/motis)) does not
+/// offer a blind-specific pedestrian profile yet.
+/// Walking routes honour the profile via more detailed route instructions.
```
CommanderStorm:
```suggestion
```

